PR #24, opened 2017-05-23
examples

frairon:


examples/2-clicks/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+Note that errors returned by the codec lead to a shutdown of the
+processor/view/emitter using it immediately. We chose that fail-early-approach since data corruption would occur if,
+for example, a processor accidentally reads and writes using a wrong codec and mixes different codecs in the group table.
+If you need to tolerate codec-errors, you'll have to handle them inside the codec and make sure it returns a `nil`-error.
+
+### Partitioning and Concurrency
+
+Messages are partitioned in Kafka using the message key. Messages within each partition are processed sequentially,
+but in parallel for all partitions. That means for our case, different users can be modified
```
SamiHiltunen: This sounds a bit wonky, maybe something like ```Messages are processed sequentially inside a given partition but concurrently between separate partitions.```

examples/2-clicks/README.md
```diff
@@ -0,0 +1,203 @@
...
+in parallel. A single user however is always modified sequentially, because all messages for that user are assigned
```
SamiHiltunen: parallel -> concurrently

examples/2-clicks/README.md
```diff
@@ -0,0 +1,203 @@
...
+to the same partition which runs sequentially.
```
SamiHiltunen: Perhaps we should add a small explanation here as to why the messages for the user are assigned to the same partition, ie. ``` Since we are using the user id as the message's key...```

frairon: done

examples/2-clicks/README.md
frairon: done

examples/2-clicks/README.md
frairon: done

PR #25, opened 2017-06-01
logger interface and options

SamiHiltunen:
* Adds options for supplying the logger as an option to processors,
  views, emitters, monitor and query. This makes custom logging possible.

* Adds a default JSON marshaler for querying and changes the
  marshaler from a required parameter to an optional one.

### DIFF
--- a/mocks.go
+++ b/mocks.go
@@ -67,7 +67,7 @@ func (cm *clientMock) GetOffset(topic string, partitionID int32, time int64) (in
 	case sarama.OffsetOldest:
 		offset, hasOffset = cm.oldestOffsets[fmt.Sprintf("%s/%d", topic, partitionID)]
 	default:
-		log.Fatalln("we don't mock this case")
+		log.Panic("we don't mock this case")
 	}
 
 	if hasOffset {
--- a/options.go
+++ b/options.go
@@ -6,7 +6,9 @@ import (
 	"time"
 
 	"github.com/lovoo/goka/kafka"
+	"github.com/lovoo/goka/logger"
 	"github.com/lovoo/goka/storage"
+
 	metrics "github.com/rcrowley/go-metrics"
 )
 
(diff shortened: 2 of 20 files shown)


mocks.go
```diff
@@ -67,7 +67,7 @@ func (cm *clientMock) GetOffset(topic string, partitionID int32, time int64) (in
 	case sarama.OffsetOldest:
 		offset, hasOffset = cm.oldestOffsets[fmt.Sprintf("%s/%d", topic, partitionID)]
 	default:
-		log.Fatalln("we don't mock this case")
+		log.Panic("we don't mock this case")
```
frairon: did you mean to change only the level here? The same logger is still used. maybe you want to add the goka-logger.

options.go
```diff
@@ -210,7 +225,7 @@ func (opt *poptions) storagePathForPartition(topic string, partitionID int32) st
 }
 
 func (opt *poptions) defaultStorageBuilder(topic string, partition int32, codec Codec, reg metrics.Registry) (storage.Storage, error) {
-	return storage.New(opt.storagePathForPartition(topic, partition), codec, reg, opt.storageSnapshotInterval)
+	return storage.New(logger.Default(), opt.storagePathForPartition(topic, partition), codec, reg, opt.storageSnapshotInterval)
```
frairon: this should use the option's logger instead of the Default one, otherwise setting a logger does not work for the storage because it always uses the default one.

options.go
```diff
@@ -348,7 +372,7 @@ func (opt *voptions) storagePathForPartition(topic string, partitionID int32) st
 }
 
 func (opt *voptions) defaultStorageBuilder(topic string, partition int32, codec Codec, reg metrics.Registry) (storage.Storage, error) {
-	return storage.New(opt.storagePathForPartition(topic, partition), codec, reg, opt.storageSnapshotInterval)
+	return storage.New(logger.Default(), opt.storagePathForPartition(topic, partition), codec, reg, opt.storageSnapshotInterval)
```
frairon: same here..

mocks.go
SamiHiltunen: Mainly just wanted to Panic here instead to output a stack trace in case this condition is hit.

options.go
SamiHiltunen: Yep, fixed that now

options.go
SamiHiltunen: fixed

mocks.go
frairon: ok sure, was just confused because you replaced the logger to use goka's logger now everywhere else, except here only the level.

PR #31, opened 2017-08-23
snapshot and batching removal, recovery transactions

### LINKED ISSUE #30: Remove snapshot from storage
The default local storage currently compacts writes with the snapshot package. The reasoning behind that was to reduce the disk traffic for some workloads. However, it seems the snapshot causes the memory usage to be way higher than expected.

The best would be to remove the write-compaction from the default storage, possibly leaving it as an optional wrapper to be passed via WithStorageBuilder().

SamiHiltunen:
* Removes snapshotting and batching from the storage. These lead
  to rather high memory usage in some cases and they can be later
  reimplemented as an additional layer if necessary. Tuning
  LevelDB configuration seems to give enough performance.

* Adds MarkRecovered method on the storage interface. This is used
  to signal that recovery has finished and that the processor/view
  can expect a lower message rate. Default storage uses this to
  disable LevelDB compaction via a transaction during recovery and
  re-enabling it afterwards.

* Removes StoragePath option. StorageBuilder now takes a complete
  storage path. Two helpers functions were added for generating
  default storage paths for views and processors.

  closes #30


SamiHiltunen:
Thanks, merging! 👍

PR #34, opened 2017-09-11
simplify context and move set offset to context commit

db7:


context.go
```diff
@@ -213,6 +214,7 @@ func (ctx *context) setValueForKey(key string, value interface{}) error {
 		return fmt.Errorf("Cannot set nil as value.")
 	}
 
+	ctx.counters.stores++
```
heltonmarx: You're increase this variable here, but probably is reading in other coprocess. For this case, it's necessary protect it with the mutex.

context.go
```diff
@@ -65,13 +65,14 @@ type context struct {
...
-		calls     int32
-		callsDone int32
+		calls     int
+		callsDone int
+		stores    int
 	}
-	m  sync.Mutex
-	wg *sync.WaitGroup
+	errors Errors
+	m      sync.Mutex
```
heltonmarx: This mutex can be substituted by a `RWMutex`, considering that you're need create a method like: 
```go
func (ctx Context) Stores() int {
	ctx.m.RLock()
	defer ctx.m.RUnlock()
	return ctx.counters.stores
}
```
The use of a `RWMutex` is recommended when for this case.

context.go
db7: Actually, it cannot/should not be accessed concurrently.

- `counters.stores` can only be incremented by the goroutine that creates the context, which is the partition-goroutine.
- promise-goroutines (coming from emit confirmations) or the partition-goroutine may read `stores` if they call `ctx.commit()`
- but only one of them will call it (the last one calling `tryCommit()`) and that happens after at least one lock-protected critical section, which will act as a memory barrier anyways.

The comment made me realize though, that `counters.calls` doesn't need a lock too. So I removed that lock and also renamed the counters.

PR #39, opened 2017-09-29
Ignore nil message

db7:


processor.go
```diff
@@ -684,21 +685,25 @@ func (g *Processor) process(msg *message, st storage.Storage, wg *sync.WaitGroup
 	// get stream subcription
 	codec := g.graph.codec(msg.Topic)
 	if codec == nil {
-		wg.Done()
```
frairon: missing `defer wg.Done()` somewhere?

db7: it's not in `ctx.finish()`

PR #42, opened 2017-10-04
Evict method for Views

SamiHiltunen:
* Adds an Evict method for Views. Evict clears the key from the
  local cache.

### DIFF
--- a/mocks.go
+++ b/mocks.go
@@ -2,10 +2,42 @@ package goka
 
 import (
 	"fmt"
+	"hash"
 	"log"
 
 	"github.com/Shopify/sarama"
 )
+
+// constHasher implements a hasher that will always return the specified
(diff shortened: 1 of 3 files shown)


mocks.go
```diff
@@ -2,11 +2,43 @@ package goka
...
+func (ch *constHasher) Reset() {}
+
+func (ch *constHasher) Size() int { return 4 }
+
+func (ch *constHasher) Write(p []byte) (n int, err error) {
+	return len(p), nil
+}
+
+func NewConstHasher(part int32) hash.Hash32 {
+	return &constHasher{partition: part}
```
frairon: if the const hasher is initialized with the partiton, shouldn't it be used then somewhere?

SamiHiltunen: good catch, fixed

PR #48, opened 2017-10-24
Expose new methods in context to set and get custom key

fsilberstein:
Hi,
Sometimes, messages we receive don't have any key.
And sometimes we listen to multiple topics with different rules for the message keys.

It would be great to be able to store value in the context via a custom key.

### DIFF
--- a/context.go
+++ b/context.go
@@ -25,8 +25,14 @@ type Context interface {
 	// Value returns the value of the key in the group table.
 	Value() interface{}
 
+	// ValueForKey returns the value of the key provided in the group table.
+	ValueForKey(key string) interface{}
+
 	// SetValue updates the value of the key in the group table.
 	SetValue(value interface{})
+
+	// SetValueForKey updates the value of the key provided in the group table.
+	SetValueForKey(key string, value interface{})
 
 	// Delete deletes a value from the group table. IMPORTANT: this deletes the
 	// value associated with the key from both the local cache and the persisted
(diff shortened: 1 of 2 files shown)


frairon:
The idea of partitioning topics using their keys is to scale horizontally, i.e. consume the events for each partition concurrently either in different processes/goroutines.
Since the messages of one partition are processed synchronously, there are no race conditions.

Allowing one partition to modify the state of another partition breaks that synchronity and would cause integrity issues like lost updates, because one partition doesn't know if another partition changed its values.
If you need to modify the value for another key, the correct way to implement that is to use the Loopback functionality of goka. 
This method sends a message to the same processor using a different key, which (possibly) ends up in another partition.

So your consume function would look like:

```
func Consume(ctx goka.Context, msg interface{}){
     // do something...

    otherValue := 42
    ctx.Loopback("other-key", otherValue)
}

func ConsumeLoop(ctx goka.Context, msg interface{}) {
    ctx.SetValue(msg)
}

```

The loopback consumer needs to be added to the group definition, e.g. like this:
The loopback needs a codec analogue to `Input` and `Persist`. 

 ```
goka.DefineGroup(
		"group-name",
		goka.Input(goka.Stream("input-stream"), newCodec(), Consume),
                goka.Loop(newLoopCodec(), ConsumeLoop),
		goka.Persist(newStateCodec()),
                .....
)
```

fsilberstein:
I understand now why you didn't do it.
I managed to do it with the loopback functionality of goka. But If my topic has two consumers (with two different applications), then my message can be processed twice by the other app with the loopback, can't it ? Or does the loopback only publish in the current group / processor ?

Thanks for your feedback anyway @frairon

db7:
The loopback sends a "self-message", a message to another partition of the *same processor group*, but not necessarily the same processor instance. 

If you have two processor groups (ie, applications with different group names), they will write into different group tables and the loopback of one group has nothing to do with the loopback of the other.

If you have two processor instances in the *same group* (say, you have started the program twice), then the partitions will be split among both instances. When you call loopback, the message is sent to another partition, possibly served by the other instance.

fsilberstein:
Ok thanks for this explanation. I close this PR :)

PR #61, opened 2017-12-06
reset stats after load finished

db7:


frairon:
one sec, I'll remove the monitoring-workaround too.

Issue #82, opened 2018-02-02
HowToX : Is it possible to iterate on View with a subset ? 


jbpin:
Hi, 
All is in the title.. I want to iterate on the view for a subset of key as with syndtr/goleveldb
 ```
iter := db.NewIterator(util.BytesPrefix([]byte("foo-")), nil)
for iter.Next() {
	// Use key/value.
	...
}
iter.Release()
err = iter.Error()
```

Can you give me some direction please ? I'm also keen to manage a PR for that if necessary.

db7:
The View iterator would does not provide that. But if you want to implement that, go ahead. You'll find the iterator implementation inside `storage`. You can combine the iterators of each partition (since each partition is a DB) with the multi iterator, also defined in the storage package. Let me know if you need some help.

db7:
Thanks for contributing. I'll close this issue since your PR #85 resolves it.

jbpin:
You are welcome.. appreciate.

PR #85, opened 2018-02-05
added support for Seek on iterator and IteratorWithRange on storage

jbpin:
As discuss in issue #82 Here is the pull request. 

I write light test on memory.. I'm looking forward for your code review.

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -9,7 +9,7 @@ type Iterator interface {
 	Key() string
 	Value() (interface{}, error)
 	Release()
-	Seek(key []byte) bool
+	Seek(key string) bool
 }
 
 type iterator struct {
@@ -43,6 +43,6 @@ func (i *iterator) Release() {
 	i.iter.Release()
 }
 
-func (i *iterator) Seek(key []byte) bool {
-	return i.iter.Seek(key)
+func (i *iterator) Seek(key string) bool {
+	return i.iter.Seek([]byte(key))
 }
--- a/iterator.go
+++ b/iterator.go
@@ -9,6 +9,7 @@ type Iterator interface {
 	Key() string
 	Value() (interface{}, error)
 	Release()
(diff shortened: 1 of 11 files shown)


iterator.go
```diff
@@ -9,6 +9,7 @@ type Iterator interface {
 	Key() string
 	Value() (interface{}, error)
 	Release()
+	Seek(key []byte) bool
```
db7: `goka.Iterator` has keys as strings. Does `Seek()` make sense when based on strings?

jbpin:
Hi guys, any update on this ?

jbpin:
Sure ;)

```
iterator.Seek([]byte('this is my key'))
```

db7:
;)

What I meant is that the whole interface of `goka.Iterator` is defined with keys as strings. So `Seek()` should take a string, like this:

```
type Iterator interface {
	Next() bool
	Key() string
	Value() (interface{}, error)
	Release()
	Seek(key string) bool
}
```

jbpin:
;) I see... In fact I added like that so that it's complient with the interface of the leveldb library. 
https://github.com/syndtr/goleveldb/blob/master/leveldb/iterator/iter.go

db7:
I guessed so. The `storage.Iterator` is defined with keys as byte slices because LevelDB uses byte slices.

Still, since we work with string keys at the goka package level, I think it would make sense to keep string keys there. Or change all keys to byte slices, but that would break way too many things.

jbpin:
ok get it.. I update the code

db7:
Excellent! And this one should take strings too: `func (v *View) IteratorWithRange(start, limit []byte) (Iterator, error) {`

Besides these `[]byte` thing, the PR looks pretty good to me. Thanks for contributing.

jbpin:
ok let me fix that

PR #88, opened 2018-02-07
messaging example

db7:
@burdiyan: can you please check the example in https://github.com/lovoo/goka/tree/example-messaging/examples/3-messaging ?

It involves most Context methods including Lookup, Join and Loopback.


burdiyan:
I'll take a look during the weekend! Thanks!

db7:
I'll merge this for now. We can always update the example later.

PR #96, opened 2018-03-01
added redis storage option

heltonmarx:
Inserted as a storage option the use of Redis, passing through Builder the [redis client](https://gopkg.in/redis.v5).

Please, any suggestions will always be welcome.

### DIFF
--- a/storage/redis/builders.go
+++ b/storage/redis/builders.go
@@ -1,0 +1,20 @@
+package redis
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/lovoo/goka/storage"
+
+	redis "gopkg.in/redis.v5"
+)
+
+// RedisBuilder builds redis storage.
+func RedisBuilder(client *redis.Client, namespace string) storage.Builder {
+	return func(topic string, partition int32) (storage.Storage, error) {
+		if namespace == "" {
+			return nil, errors.New("missing namespace to redis storage")
+		}
+		return New(client, fmt.Sprintf("%s:%s:%d", namespace, topic, partition))
+	}
+}
--- a/storage/redis/redis.go
+++ b/storage/redis/redis.go
@@ -1,21 +1,26 @@
-package storage
+package redis
 
 import (
 	"errors"
 	"fmt"
 	"strconv"
-	"time"
+
+	"github.com/lovoo/goka/storage"
 
 	redis "gopkg.in/redis.v5"
 )
 
+const (
+	offsetKey = "__offset"
+)
+
 type redisStorage struct {
-	client    *redis.Client
-	retention time.Duration
+	client *redis.Client
+	hash   string
 }
 
-// NewRedis creates a new Storage backed by Redis.
-func NewRedis(client *redis.Client, retention time.Duration) (Storage, error) {
+// New creates a new Storage backed by Redis.
+func New(client *redis.Client, hash string) (storage.Storage, error) {
 	if client == nil {
 		return nil, errors.New("invalid redis client")
 	}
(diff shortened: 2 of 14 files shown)


db7:
Really nice contribution! Thanks for taking the time.

I have a couple of comments:

- I think it would be better to create a subpackage `storage/redis` and put this code there, otherwise the Redis client will become a default dependency.
- the keys stored in Redis need to be made disjoint from other possible applications writing to the same Redis instance. I'd suggest prefixing keys with something like `namespace:table:partition:`.
    - namespace should distinguish applications that write to the same keys
    - table should distinguish multiple tables in the same application that write to the same keys
    - partition should guarantee that partitions don't overwrite each other. Actually that is already the case due to the hasher that splits the key space in partitions, but all partitions have at least one key in common called `__offset_key`, which stores the latest offset of the partition for recovery. If this gets overwritten, partitions will be recovered incorrectly.
    - You can pass these options in the builder.

And I saw that you have added a retention to the entries in Redis. What is the intention of expiring the entries? Is this to be used in combination with a retention in Kafka too? I think one should have the option of setting no retention (it seems to be mandatory now).

storage/redis/builders.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/lovoo/goka/storage"
+
+	redis "gopkg.in/redis.v5"
+)
+
+// RedisBuilder builds redis storage.
+func RedisBuilder(client *redis.Client, namespace string) storage.Builder {
+	return func(topic string, partition int32) (storage.Storage, error) {
+		if namespace == "" {
+			namespace = strconv.FormatInt(time.Now().UnixNano(), 10)
```
SamiHiltunen: It would be better to return an error here and make the namespace a required parameter. Using the time as the namespace would mean that the application has to recover every launch and leave a lot of obsolete data from previous runs in Redis.

storage/redis/redis.go
```diff
@@ -0,0 +1,183 @@
...
+	offsetKey = "__offset"
+)
+
+type redisStorage struct {
+	client *redis.Client
+	hash   string
+}
+
+// NewRedis creates a new Storage backed by Redis.
+func NewRedis(client *redis.Client, hash string) (storage.Storage, error) {
```
SamiHiltunen: I would remove the Redis suffixes and prefixes from the types and methods as the package path and name implies this is a Redis storage for Goka. I think it would be cleaner to have `redis.New`.

storage/redis/redis.go
```diff
@@ -0,0 +1,183 @@
...
+	return nil
+}
+
+func (s *redisStorage) Open() error {
+	return nil
+}
+
+// Close removes all data present current hash(`namespace:table:partition`)
+// instead closing the redis client.
+func (s *redisStorage) Close() error {
```
SamiHiltunen: `Close` should close the client and leave the data in Redis. Deleting the data on close would force the applications to recover their state on every launch and this is not in line with the other storage implementations. Functionality for cleaning the state from Redis would definitely be a convenient addition but I would move it to a separate method.

db7: I don't know why it is deleting the data, but I don't think that `Close` should actually close the client too. The caller should care about closing the client since the Redis client is passed to the storage builder and all partitions share the same client.

SamiHiltunen: I completely overlooked that part. Even if it wasn't shared with other builders, it would be better to leave it to caller who built the client.

heltonmarx: I put this `delete` here, 'cause I thought that when you close a partition, you would not use it anymore by this consumer.

PR #105, opened 2018-03-13
add restartable view support

### LINKED ISSUE #102: View should have option to fallback to local-only mode
In some cases, one wants to access the view contents even if connecting to Kafka is not possible or broken. We could add an option to views so that they don't close the databases when an error in the Kafka connection occurs. @sweigert

db7:
This PR resolves #102.

If one starts a view with the option`goka.WithViewRestartable()`, the view may be restarted or used, eg, by calling `Get()` and `Recovered()`, `Iterator()`, etc. To release all resources, ie, close the underlying local storage, one must call `Stop()` when using the `Restartable()` option.

### DIFF
--- a/view.go
+++ b/view.go
@@ -16,6 +16,7 @@ type Getter func(string) (interface{}, error)
 
 // View is a materialized (i.e. persistent) cache of a group table.
 type View struct {
+	brokers    []string
 	topic      string
 	opts       *voptions
 	partitions []*partition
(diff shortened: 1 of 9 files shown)


view.go
```diff
@@ -110,6 +112,12 @@ func (v *View) createPartitions(brokers []string) (err error) {
 
 // reinit (re)initializes the view and its partitions to connect to Kafka
 func (v *View) reinit() error {
+	v.mInit.Lock()
+	if v.terminated {
+		return fmt.Errorf("view: cannot reinitialize terminated view")
+	}
+	defer v.mInit.Unlock()
```
SamiHiltunen: This defer should be above the conditional

db7:
True, I just fixed it.

view.go
```diff
@@ -166,14 +198,39 @@ func (v *View) stop() {
...
+
+// Stop stops the view, closes storage partitions, and frees resources
 func (v *View) Stop() {
 	v.opts.log.Printf("View: stopping")
+
+	// do not allow any reinitialization
+	v.mInit.Lock()
+	if v.terminated {
+		return
+		v.mInit.Unlock()
```
tsne: This unlock is dead code and could result in a deadlock.

db7: damn it. Thanks

PR #118, opened 2018-03-27
Feature/move kafkamock

db7:
This PR refactors KafkaMock a bit.

- KafkaMock is renamed as Tester since it mocks not only Kafka but also the storage
- Tester is moved into tester subpackage
- Goka accepts a tester interface as option in processors (that should be extended later for views too)
- The ControllerMock in `kafkamock.go` is now in `mock/controller.go`
- The old Tester type in `kafkamock.go` is not `T` since it wraps `testing.T`
- The constructor of the new Tester does not require the group since that is passed to the consumer builder

### DIFF
--- a/examples/testing/main_test.go
+++ b/examples/testing/main_test.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/golang/mock/gomock"
 	"github.com/lovoo/goka"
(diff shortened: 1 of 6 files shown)


frairon:
I added another test in the examples to check for emits.
@db7 please recheck.

PR #139, opened 2018-08-03
Fix passing opts to LevelDB in BuilderWithOptions

mhaitjema:
I believe I found a bug in the `storage` package in `BuilderWithOptions` where the options are never propagated to leveldb as the `opts` arg is currently unused but should in the call to `OpenFile`.


db7:
@mhaitjema good catch! Thanks for the PR.

Issue #145, opened 2018-09-27
Improve tester implementation


frairon:
The current implementation of goka tester has some limitations:
* only one processor can be tested
* lookups and joins are not supported
* state codec has to be registered manually

In branch `tester-refactoring` I started writing a new tester implementation. 
The new tester features
* multiple processor tests
* lookups, joins and loop fully working
* group graph detection, so every topic knows its codec
* synchronous, so after `tester.Consume(..)` has terminated, all processors and loops and are finished consuming.

The interface will change slightly, assuming that the tester has not been used too much yet:

Starting and running the tester has not changed.

## Changes 

* retrieving values from the processor's group table via the tester was done before via
```
tester.ValueForKey("somekey")
```
This was possible because there was only one group table. Now we can have multiple, so you have to specify the topic too
```
tester.TableValue("group-table", "somekey")
```
* Setting a value has changed analogue like
```
tester.SetValue("somekey", "value")
// becomes
tester.SetTableValue("group-table", "somekey", "value")
```

* the ExpectEmit functionality was cumbersome because the tester didn't know a topic's codec:
```
tester.ExpectEmit("topic", "key", func(value []byte){
   // decode the value and check what you need to check
})

// then to make sure nothing extra was emitted
tester.Finish(true)
```
Now we can do
``` 
key, value, hasNext := tester.NextMessage("topic")
if !hasNext{
  // there was nothing more in this topic
}else{
    // check what you need to check with key and value
}
// similar to check there were no more messages
tester.ExpectEmpty("topic")
```
The old ExpectEmit is still there to avoid too much broken tests when upgrading but imho it should be removed.

Documentation and Examples still need lots of polishing, but that's essentially how the new Tester would look like.
What do you think?

PR #146, opened 2018-10-01
improve tester implementation

frairon:
Reimplementation of the tester to  fix all issues (see issue #145)


examples/4-tests/example_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// consume a message
+	gkt.ConsumeString("input", "key", "some message")
+
+	// ensure the message was received
+	ensure.DeepEqual(t, receivedMessage, "some message")
+}
+
+// Scenario (2)
+//
```
juandes: Silly comment. You have this unused comment line :)

tester/messagetracker.go
```diff
@@ -0,0 +1,95 @@
+package tester
+
+// MessageTracker tracks message offsets for each topic for convenient
+// 'expect message x to be in topic y' in unit tests
+type MessageTracker struct {
```
db7: I would suggest bind the MessageTracker to a topic. That will simplify the interface.

tester/messagetracker.go
```diff
@@ -0,0 +1,95 @@
...
+	return &MessageTracker{
+		t:                t,
+		nextMessageIndex: make(map[string]int),
+		tester:           tester,
+	}
+}
+
+// NextMessage returns the next message from the topic since the last time this
+// function was called (or MoveToEnd)
+func (mt *MessageTracker) NextMessage(topic string) (string, interface{}, bool) {
```
db7: suggestion: use `Next()` instead

tester/messagetracker.go
```diff
@@ -0,0 +1,95 @@
...
+
+	decoded, err := mt.tester.codecForTopic(topic).Decode(msgRaw)
+	if err != nil {
+		mt.t.Fatalf("Error decoding message: %v", err)
+	}
+	return key, decoded, true
+}
+
+// NextMessageRaw returns the next message in passed topic
+func (mt *MessageTracker) NextMessageRaw(topic string) (string, []byte, bool) {
```
db7: suggestion: use `NextRaw()` instead

tester/messagetracker.go
```diff
@@ -0,0 +1,95 @@
...
+		return "", nil, false
+	}
+	msg := q.message(mt.nextMessageIndex[topic])
+
+	mt.nextMessageIndex[topic]++
+	return msg.key, msg.value, true
+}
+
+// ExpectEmpty ensures the topic does not contain more messages
+func (mt *MessageTracker) ExpectEmpty(topic string) {
```
db7: This is not necessary since once can call `Next()` until it returns false

tester/messagetracker.go
```diff
@@ -0,0 +1,95 @@
...
+func (mt *MessageTracker) MoveAllToEnd() {
+	for topic := range mt.tester.topicQueues {
+		mt.nextMessageIndex[topic] = int(mt.tester.queueForTopic(topic).hwm)
+	}
+}
+
+// ExpectEmit ensures a message exists in passed topic and key. The message may be
+// inspected/unmarshalled by a passed expecter function.
+// DEPRECATED: This function is only to get some compatibility and should be removed in future
+func (mt *MessageTracker) ExpectEmit(topic string, key string, expecter func(value []byte)) {
```
db7: I'd remove this method.

tester/tester.go
```diff
@@ -20,35 +23,84 @@ type Codec interface {
...
+	q, exists := km.topicQueues[topic]
+	if !exists {
+		panic(fmt.Errorf("No queue for topic %s", topic))
+	}
+	return q
+}
+
+// CreateMessageTracker creates a message tracker that starts tracking
+// the messages from the end of the current queues
+func (km *Tester) NewMessageTrackerFromEnd() *MessageTracker {
```
db7: suggestion: `NewMessageTracker(topic string)`

tester/messagetracker.go
frairon: actually I'd leave the message, as it does more than just checking we're at the end. Also having a shortcut for the alternative loop that we would have to replae in all projects would be helpful.
but I'd rename it to `ExpectAtEnd` to have it consistend with `MoveToEnd`.
What do you think?

db7: What about replacing `ExpectAtEnd()` with something like `mt.Offset() == mt.HWM()` ?

and  `MoveToEnd()` replacing with `mt.Seek(mt.HWM())`?

frairon: agreed

PR #165, opened 2018-12-14
terminate tester when one of the queue consumers died

frairon:
unit tests using `tester.Tester` would not terminate, whenever there's a panic  in the tested processor.
That PR fixes it.

### DIFF
--- a/tester/queue.go
+++ b/tester/queue.go
@@ -2,6 +2,7 @@ package tester
 
 import (
 	"fmt"
+	"log"
 	"sync"
 )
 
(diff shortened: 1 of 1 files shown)


tester/queue.go
```diff
@@ -82,15 +83,21 @@ func (q *queue) waitConsumersInit() {
...
+			logger.Printf(" --> %s is running", cons.queue.topic)
+		}
 	}
 
 	for cons := range q.simpleConsumers {
 		logger.Printf("waiting for simple consumer %s to be ready", cons.queue.topic)
 		select {
 		case <-cons.state.WaitForState(running):
 		case <-cons.state.WaitForState(stopped):
+		case <-cons.state.WaitForState(killed):
```
db7: a bit weird to be ready when it's killed.

frairon: okay. any suggestions how to improve?

Issue #176, opened 2019-03-08
processor gets stuck in shutdown after error


frairon:
We have seen cases where a running processor gets stuck while shutting down, when one of the brokers goes down or the metadata is changed due to e.g. rebalance.
The error is supposedly a `kafka.Promise` that is never finished, because the producer gets cancelled. This leads to the waitgroup in [https://github.com/lovoo/goka/blob/master/partition.go#L142](partition.go) to never get done, so the partition never shuts down, thus the whole processor gets stuck.
We could not reproduce the bug, but a stack trace of one of the stuck processors showed this location.

PR is open, hope that will fix the issue.

db7:
The producer does not get canceled on rebalance. I think it should be then related to some error on emitting a message. Do you have the stack trace?

frairon:
Sorry, my bad. The example was just wrong. So one broker goes down -> the leader of some partitions is not the leader anymore -> the meatdata is changed. I'll send you the stack trace

PR #177, opened 2019-03-08
#176 bugfix processor shutdown when producer gets closed

frairon:


frairon:
The issue is still appearing with this fix, so I'll close this

PR #180, opened 2019-04-02
178 multi_iterator looks broken

frairon:


storage/multi_iterator.go
```diff
@@ -36,18 +44,24 @@ func (m *multiIterator) Release() {
...
+	m.iters = nil
+
+	// at least one iterator's seek was successful
 	ok := false
-	for i := range m.iters {
-		if m.iters[i].Seek(key) {
-			iters = append(iters, m.iters[i])
+
+	// iterate over all (original) iterators
+	for _, iter := range m.origIters {
```
ubunatic: You could omit the `origIters` in the struct and cache the old iters here: `origIters := m.iters`. Then set `m.iters = nil` and loop over the local `origIters` to rebuild the `m.iters`.

This should work unless you allow `Seek` after a `Release`

PR #183, opened 2019-04-18
start logging while recovering

TheL1ne:


processor.go
```diff
@@ -523,10 +523,27 @@ func (g *Processor) newStorage(topic string, id int32, update UpdateCallback) (*
...
 	}
 
-	st, err := g.opts.builders.storage(topic, id)
+	wg := new(sync.WaitGroup)
+	var err error
+	var st storage.Storage
+	ticker := time.NewTicker(5 * time.Minute)
+	defer ticker.Stop()
+	go func() {
+		wg.Add(1)
```
ubunatic: Add(1) before starting the goroutine. Otherwise the Wait() at the end may be called before the Add(1)

frairon: this wg.Add(1) needs to be outside of the go-routine, otherwise you might be super super unlucky and the outer go-routine might come to the wg.Wait()-call before the go-routine has started and won't wait then.

processor.go
```diff
@@ -523,10 +523,27 @@ func (g *Processor) newStorage(topic string, id int32, update UpdateCallback) (*
 		}, nil
 	}
 
-	st, err := g.opts.builders.storage(topic, id)
+	wg := new(sync.WaitGroup)
```
frairon: Don't know why, but usually I'd create a WaitGroup not as a pointer, so doing `var wg sync.WaitGroup` instead.
But that's just personal flavor. Either way you could then use the multiline-var with `var ( ... ) ` for the other variables below too.

processor.go
```diff
@@ -523,10 +523,27 @@ func (g *Processor) newStorage(topic string, id int32, update UpdateCallback) (*
...
+	var err error
+	var st storage.Storage
+	ticker := time.NewTicker(5 * time.Minute)
```
frairon: let's have a smaller ticker, like every 10 seconds? We wanted to avoid having to wait too long until we see anything

processor.go
```diff
@@ -523,10 +523,27 @@ func (g *Processor) newStorage(topic string, id int32, update UpdateCallback) (*
...
+	defer ticker.Stop()
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+		st, err = g.opts.builders.storage(topic, id)
+		g.opts.log.Printf("finished building storage for topic %s", topic)
+	}()
+	go func() {
+		for range ticker.C {
+			g.opts.log.Printf("building storage for topic %s ...", topic)
```
frairon: It would be nice if you could add some information how long we have been waiting (like `... building storage for topic x for y seconds`)

processor.go
db7: I agree, don't allocate the wait group if its lifetime is only within this function scope. That would only create confusion for those reading the code.

processor.go
TheL1ne: But because this takes hours it would pollute the log thats why I used a rather big ticker.

PR #199, opened 2019-08-08
Tester tolerate unconsumed queues

TheL1ne:
I rebased this branch onto the master. Hopefully I did not do any mistakes there @frairon.
Tests still work fine.

I want this in the master because we have version conflicts between projects otherwise.

### DIFF
--- a/integrationtest/processor_stuck.go
+++ b/integrationtest/processor_stuck.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"sync/atomic"
+	"time"
+
+	"github.com/lovoo/goka"
+	"github.com/lovoo/goka/codec"
+	"github.com/lovoo/goka/kafka"
+)
+
+func main() {
+
+	proc, err := goka.NewProcessor([]string{"localhost:9092"},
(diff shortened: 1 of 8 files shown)


view.go
```diff
@@ -139,13 +138,13 @@ func (v *View) Run(ctx context.Context) error {
 	for id, p := range v.partitions {
 		pid, par := int32(id), p
 		errg.Go(func() error {
-			v.opts.log.Printf("view: partition %d started", pid)
+			v.opts.log.Printf("view [%s]: partition %d started", pid)
```
frairon: @TheL1ne  while you're on it, could you add `v.Topic()` as first value here?

integrationtest/processor_stuck.go
```diff
@@ -0,0 +1,121 @@
...
+
+		log.Printf("Running processor")
+		procRunErr := proc.Run(ctx)
+		log.Printf("Processor finished with %v", procRunErr)
+	}()
+
+	log.Printf("wait 5 seconds before starting to emit")
+	time.Sleep(5 * time.Second)
+
+	for i := 0; i < 50; i++ {
```
175-kavinski: These loops could use some more overview imo

PR #200, opened 2019-08-23
protect storage writes

TheL1ne:
We are creating multiple processors in a short time in our tests and they are failing sometimes because of concurrent map writes. We think we traced the problem to this point.

### DIFF
--- a/storage/memory.go
+++ b/storage/memory.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/syndtr/goleveldb/leveldb/util"
 )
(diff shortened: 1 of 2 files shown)


ubunatic:
> Protecting that makes sense absolutely, but if you only protect that one code branch we might only buy some time and the problem will reappear later. So I suggest using the lock everywhere mStorages is used. Use `RLock/RUnlock` when reading only and `Lock/Unlock` for writing.

Sounds great!
@TheL1ne Please implement it as proposed.

TheL1ne:
@frairon did you meant it the way it is now?

I also did a similar thing for the in-memory storage defined outside of the tester because I think a similar problem could occur.

PR #206, opened 2019-10-22
adding custom logger option

175-kavinski:
Added the ability to use a custom logger implemented according to the goka logger interface.
This shouldn't break the current usage, since functions with the default logger stay the same.


175-kavinski:
May be not needed due to the options

Issue #207, opened 2019-11-06
Dropping message while rebalancing


WideLee:
I have tested on 2-clicks example by running **one producer** to emit user's clicks and **two consumers** to statistics clicks. the message emitted by producer such as follow:

```
user-0  0-<time>
user-1  1-<time>
user-2  2-<time>
user-3  3-<time>
user-4 4-<time>
user-5  5-<time>
user-6  6-<time>
user-7  7-<time>
user-8  8-<time>
user-9  9-<time>
user-0  10-<time>
user-1  11-<time>
user-2  12-<time>
...
user-8  18-<time>
user-9  19-<time>
user-0  20-<time>
```

However, after the second consumer started and kafka rebalancing, the clicks count of first consumer is smaller than expected, the missing count equals to the number of the message "dropping message from topic = user-click while loading".

```
[proc] key: user-0 clicks: 16, msg: 150-2019-11-06 23:32:01.300978 +0800 CST m=+16.488060985
[proc] key: user-1 clicks: 16, msg: 151-2019-11-06 23:32:01.4099 +0800 CST m=+16.596983868
[proc] key: user-2 clicks: 16, msg: 152-2019-11-06 23:32:01.518171 +0800 CST m=+16.705255671
[proc] key: user-3 clicks: 16, msg: 153-2019-11-06 23:32:01.628471 +0800 CST m=+16.815555953
[proc] key: user-4 clicks: 16, msg: 154-2019-11-06 23:32:01.73763 +0800 CST m=+16.924715895
[proc] key: user-5 clicks: 16, msg: 155-2019-11-06 23:32:01.847644 +0800 CST m=+17.034731167
[proc] key: user-6 clicks: 16, msg: 156-2019-11-06 23:32:01.957907 +0800 CST m=+17.144994713
[proc] key: user-7 clicks: 16, msg: 157-2019-11-06 23:32:02.066497 +0800 CST m=+17.253585582
[proc] key: user-8 clicks: 16, msg: 158-2019-11-06 23:32:02.175928 +0800 CST m=+17.363017306
2019/11/06 23:32:02 Processor: dispatcher stopped
2019/11/06 23:32:02 partition mini-group-15-table/0: exit
2019/11/06 23:32:02 partition mini-group-15-table/1: exit
2019/11/06 23:32:02 partition mini-group-15-table/2: exit
2019/11/06 23:32:02 partition mini-group-15-table/3: exit
2019/11/06 23:32:02 Removing partition 1
[...]

PR #208, opened 2019-11-09
repeating messages from input stream after recover

WideLee:
Fix the issue in  [ISSUE 207] (https://github.com/lovoo/goka/issues/207)
Some messages from input stream are dropped without processing after rebalancing.

### DIFF
--- a/partition.go
+++ b/partition.go
@@ -57,6 +57,8 @@ type partition struct {
 	lastStats     *PartitionStats
 	requestStats  chan bool
 	responseStats chan *PartitionStats
+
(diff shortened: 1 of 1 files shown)


frairon:
Thanks very much @WideLee for the PR!
It looks good so far, but we'd like to reproduce the bug and the fix first with a newer version of Kafka. We're setting up a new cluster in the next couple of days. Give us some days to check that and then I'm happy to merge :).

Cheers!

frairon:
Sorry @WideLee, we're not there yet :(

PR #214, opened 2019-12-30
Adding header on message

toninho09:
I changed the message to contain the headers

only work if set the `client.Version` to 0.11+ on samara client 

Example processor with `config.Version` 2.2.0
```
p, err := goka.NewProcessor(
	brokers,
	g,
	goka.WithConsumerBuilder(func(brokers []string, group, clientID string) (consumer kafka.Consumer, err error) {
		config := kafka.NewConfig()
		config.ClientID = clientID
		config.Version = sarama.V2_2_0_0
		return kafka.NewSaramaConsumer(brokers, group, config)
	}),
)
```

Your can get the headers from context with `ctx.Headers()`

### DIFF
--- a/context.go
+++ b/context.go
@@ -33,6 +33,9 @@ type Context interface {
 	// Value returns the value of the key in the group table.
 	Value() interface{}
 
+	// Headers returns the headers of the input message
+	Headers() map[string][]byte
+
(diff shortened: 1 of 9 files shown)


db7:
That's nice! LGTM @frairon

PR #221, opened 2020-01-24
Test partition_table

R053NR07:
move mocks
add consumer mock


mockconsumer.go
```diff
@@ -0,0 +1,327 @@
...
+// message was consumed from the Messages channel, because there are legitimate
+// reasons forthis not to happen. ou can call ExpectMessagesDrainedOnClose so it will
+// verify that the channel is empty on close.
+func (pc *MockPartitionConsumer) YieldMessage(msg *sarama.ConsumerMessage) {
+	pc.l.Lock()
+	defer pc.l.Unlock()
+
+	msg.Topic = pc.topic
+	msg.Partition = pc.partition
+	msg.Offset = atomic.AddInt64(&pc.highWaterMarkOffset, 1)
```
frairon: you'll skip message with offset 0, because the first one will have offset 1 (sync.AddInt64 returns the value after the increment).
Probably do a `-1` here and avoid the `+1` in the getter function `HighWaterMarkOffset()`. That also makes the variable correct. Currently it's the last message's offset, not the hwm.

partition_table.go
```diff
@@ -239,21 +251,24 @@ func (p *PartitionTable) load(ctx context.Context, stopAfterCatchup bool) (rerr
 		return
 	}
 
+	p.log.Printf("Offset stored locally: %d", localOffset)
```
frairon: remove print

partition_table.go
```diff
@@ -239,21 +251,24 @@ func (p *PartitionTable) load(ctx context.Context, stopAfterCatchup bool) (rerr
...
+
 	loadOffset, hwm, err := p.findOffsetToLoad(localOffset)
 	if err != nil {
 		errs.Collect(err)
 		return
 	}
 
 	if localOffset >= hwm {
-		errs.Collect(fmt.Errorf("local offset is higher than partition offset. topic %s, partition %d, hwm %d, local offset %d", p.topic, p.partition, partitionHwm, localOffset))
+		errs.Collect(fmt.Errorf("local offset is higher than partition offset. topic %s, partition %d, hwm %d, local offset %d", p.topic, p.partition, hwm, localOffset))
```
frairon: let's add an extra step how to fix this, like `Probably the kafka topic was recreated but the processor still has it's old local cache`

processor_test.go
```diff
@@ -26,35 +26,66 @@ func PanicStringContains(t *testing.T, s string) {
 	}
 }
 
+func createMockBuilder(t *testing.T) (*gomock.Controller, *builderMock) {
+	ctrl := NewMockController(t)
+	bm := newBuilderMock(ctrl)
+	bm.st = storage.NewMemory()
```
frairon: I think this should be default in the buildermock, otherwise test users wil panic if they forget to add this step

PR #238, opened 2020-03-19
go mod support add

oneslash:
I would like to add go mod so it would be easier to develop


frairon:
Hi @oneslash,
what a timing... So we were working on quite a big refactoring the last weeks/months of goka, described [here](https://github.com/lovoo/goka/issues/239). 
That includes of course the migration to go mod, as well as several other improvements and changes. 
Anyway, many thanks for your contribution!
However, to avoid further conflicts when we finally want to merge the refactoring, I'd prefer not to merge the PR right now. We'll tag the branch with go mod and the refactoring hopefully today :)

PR #243, opened 2020-04-07
added lots of godoc, fixed many linter errors, added Open call when c…

frairon:
* added lots of godoc
* fixed many linter errors
* added Open call when creating storage
* changed Iterator interface: Err() -> Error()


iterator.go
```diff
@@ -45,8 +53,8 @@ func (i *iterator) Value() (interface{}, error) {
 }
 
 // Err returns the possible iteration error.
```
R053NR07, line 55: Comment does not fit anymore.

frairon: done, renamed it back.

PR #246, opened 2020-04-15
Add Backoff when restartable CatchupForever is executed

R053NR07:
A customizable backoff is used when restarting the PartitionTables loading function. 
Options for the view and processor can be used to customize the reset timeout and set the backoff provider.

### DIFF
--- a/builders.go
+++ b/builders.go
@@ -2,6 +2,7 @@ package goka
 
 import (
 	"hash"
+	"time"
 
 	"github.com/Shopify/sarama"
 )
@@ -87,4 +88,12 @@ func SaramaConsumerBuilderWithConfig(config *sarama.Config) SaramaConsumerBuilde
 		config.ClientID = clientID
 		return sarama.NewConsumer(brokers, config)
 	}
+}
+
+// BackoffBuilder creates a backoff
+type BackoffBuilder func() (Backoff, error)
+
+// DefaultBackoffBuilder returnes a simpleBackoff with 10 second steps
+func DefaultBackoffBuilder() (Backoff, error) {
+	return NewSimpleBackoff(time.Second * 10), nil
 }
--- a/partition_processor.go
+++ b/partition_processor.go
@@ -69,7 +69,9 @@ func newPartitionProcessor(partition int32,
 	lookupTables map[string]*View,
 	consumer sarama.Consumer,
 	producer Producer,
-	tmgr TopicManager) *PartitionProcessor {
+	tmgr TopicManager,
+	backoff Backoff,
+	backoffResetTime time.Duration) *PartitionProcessor {
 
 	// collect all topics I am responsible for
 	topicMap := make(map[string]bool)
@@ -136,6 +138,8 @@ func newPartitionProcessor(partition int32,
 			opts.updateCallback,
 			opts.builders.storage,
 			log.Prefix("PartTable"),
+			backoff,
+			backoffResetTime,
 		)
 	}
 	return partProc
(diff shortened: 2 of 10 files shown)


R053NR07:
@mattburman

mattburman:
Looks good, it looks like I should be able to replicate our existing behaviour using https://github.com/jpillora/backoff

Presumably I'd just implement the goka.Backoff interface to use the existing library. We wouldn't need it to reset however, so I suppose we can just implement an empty function and not pass in the `WithViewBackoffResetTimeout`?

mattburman:
Ignore the above, I misunderstood the purpose of `WithViewBackoffResetTimeout`. We would need to implement Reset so when the view reconnects, it will time out and then call reset so we can reset our backoff.

partition_processor.go
```diff
@@ -69,7 +69,9 @@ func newPartitionProcessor(partition int32,
 	lookupTables map[string]*View,
 	consumer sarama.Consumer,
 	producer Producer,
-	tmgr TopicManager) *PartitionProcessor {
+	tmgr TopicManager,
+	backoff Backoff,
```
frairon, line 73: couldn't we take that from the options instead of adding extra parameters?

processor.go
```diff
@@ -607,7 +607,11 @@ func (g *Processor) createPartitionProcessor(ctx context.Context, partition int3
 		return fmt.Errorf("processor [%s]: partition %d already exists", g.graph.Group(), partition)
 	}
 
-	pproc := newPartitionProcessor(partition, g.graph, session, g.log, g.opts, g.lookupTables, g.saramaConsumer, g.producer, g.tmgr)
+	backoff, err := g.opts.builders.backoff()
```
frairon, line 610: this could be done in the partition_processor as it has the options anyway.

builders.go
```diff
@@ -88,3 +89,11 @@ func SaramaConsumerBuilderWithConfig(config *sarama.Config) SaramaConsumerBuilde
 		return sarama.NewConsumer(brokers, config)
 	}
 }
+
+// BackoffBuilder creates a backoff
+type BackoffBuilder func() (Backoff, error)
+
+// DefaultBackoffBuilder returnes a simpleBackoff with 10 second steps
+func DefaultBackoffBuilder() (Backoff, error) {
+	return newSimpleBackoff(time.Second * 10), nil
```
frairon, line 98: If we export it like `NewSimpleBackoff`, then users could reuse the backoff and change the step duration without recreating a new backoff strategy.

partition_processor.go
R053NR07: Then the function needs to be adopted since the builder in the options can return an error and we can't ensure the backoff creation works correctly.

processor.go
R053NR07: As said above. Since this function retrunes an error this would mean altering the newPartitionProcessor signature.

builders.go
R053NR07: Thats true.

R053NR07:
> Ignore the above, I misunderstood the purpose of `WithViewBackoffResetTimeout`. We would need to implement Reset so when the view reconnects, it will time out and then call reset so we can reset our backoff.

Yes that was exactly what I intended :) But in the other case you described above I would wrap the actual backoff and add an empty reset function doing nothing. That way the reset would have been disabled.

Also I would like to add that github.com/jpillora/backoff already implements the Backoff interface. So you just have to hand it over as option and use it without further adoption.

mattburman:
> Also I would like to add that github.com/jpillora/backoff already implements the Backoff interface. So you just have to hand it over as option and use it without further adoption.

Ah I didn't realise that at first, nice one!

partition_processor.go
frairon: well good point, let's keep it then.

processor.go
frairon: true

PR #252, opened 2020-04-26
Panic documentation & Migration guide

frairon:


MIGRATION.md
```diff
@@ -0,0 +1,41 @@
...
+    // handle error
+
+    // sleep or simple backoff logic
+    time.Sleep(time.Second)
+  }
+}()
+
+// After migration:
+// create a view
+view09x, _ := NewView(..., WithAutoReconnect())
```
mattburman, line 29: should this be `WithViewAutoReconnect`?

MIGRATION.md
```diff
@@ -0,0 +1,41 @@
...
+In 0.1.4, if a view was created to be restartable, it returned from `Run` in case of errors, but allowed
+to be restarted calling `Run` again. The view was still usable, even if it was not running and receiving updates.
+
+The behavior of that option has changed in 0.9.x in a way that the `view.Run` does the reconnect internally using configurable backoff.
+The Option was also renamed (the old version has been kept for compatibility reasons).
+
+```go
+
+// create a view
+view014, _ := NewView(..., WithRestartable())
```
mattburman, line 14: Similarly this is `WithViewRestartable`. Though maybe it's good to be "incorrect" since it's the old version and shouldn't be copied since it's just an example

mattburman:
This is really clear, nice work 👍

MIGRATION.md
```diff
@@ -0,0 +1,41 @@
+This document sums up issues for migrating between 0.1.4 to 0.9.x
```
mattburman, line 1: The guide is really clear, nice work 👍

MIGRATION.md
frairon: good catch Matt. Should've tried to compile the snippet I guess.

MIGRATION.md
frairon: fixed it

MIGRATION.md
frairon: If there something else that needs to be documented let me know. It's just a start...

PR #253, opened 2020-04-26
Add boltdb storage option

p4tin:
Not sure if that was the right thing but I modeled the bold storage with the refis one.  As well I think we can also try `Badger DB (https://github.com/dgraph-io/badger)` which may be more on par when it comes to write/read key speed vs leveldb and also rocksdb.

### DIFF
--- a/storage/bolt/bolt.go
+++ b/storage/bolt/bolt.go
@@ -1,0 +1,236 @@
+package bolt
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/boltdb/bolt"
+
+	"github.com/lovoo/goka/storage"
+)
+
+const (
+	offsetKey = "__offset"
+)
+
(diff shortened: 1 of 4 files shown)


storage/bolt/bolt.go
```diff
@@ -0,0 +1,236 @@
...
+
+	value, err := strconv.ParseInt(string(data), 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("error decoding bolt offset (%s): %v", string(data), err)
+	}
+	return value, nil
+}
+
+func (b *boltStorage) Set(key string, value []byte) error {
+	err := b.client.Update(func(tx *bolt.Tx) error {
```
frairon, line 100: Probably we need to distinguish between recovering mode and running mode, otherwise it'll create a new transaction for every new value in recovery, which probably slows it down. But let's optimize later.

Issue #239, opened 2020-03-19
Sarama ConsumerGroup Update


frairon:
In this issue we want to announce the upcoming release of a new and majorly refactored version of Goka.
Spoiler: there will be breaking changes, but nothing that can't be fixed with simple text-replace in your projects. So no worries. We'll provide a small migration guide, but most of the things will be obvious anyway.

So what are the major changes:

* Removed sarama-cluster, replaced with sarama's ConsumerGroup
* remove package kafka, move remaining files to root
* introduce go mod
* Sorry, removed support for librdkafka

We'll tag the version shortly so it can be tested. Let us know if there are any issues so we can quickly react.

mattburman:
Hi @frairon, I've been evaluating the latest beta changes and writing a wrapper around it for our own usage. Previously we had some logic around view.Run which implemented retry behaviour with an exponential backoff.

Re-testing this logic with the new versions shows that view.Run no longer returns like before and it's reconnecting with its own logic. That's great that the retries are within Goka now and we can remove some of our own logic.

I think I've narrowed it down to view.Run now looping over each partition and calling CatchupForever: https://github.com/lovoo/goka/compare/v0.1.4...v0.9.0-beta1#diff-6202d9bde3a4cbd1c7a28821646249d4

This is great, but it would be good to have some kind of control over the retry behaviour. We previously has view.Run wrapped with a configurable backoff using https://github.com/jpillora/backoff. What are your thoughts on this? For now I am OK with removing our custom logic and keeping the default retry behaviour implemented in Goka. I would still welcome any thoughts on making the behaviour configurable, however.

I would suggest any upgrade/integration documentation includes information about this change to view.Run because it took me a while to realise our previous backoff behaviour was no longer running.

mattburman:
I have also found the view.Recovered method returns true whilst it is struggling to connect to kafka. Is this intended?

mattburman:
> I have also found the view.Recovered method returns true whilst it is struggling to connect to kafka. Is this intended? I need some way of determining whether the app is connected to kafka for metrics. It looks like partitionStats.Status also returns 4 (PartitionRunning) when no connection to kafka can be established. Stalled returns false too

Looks like #246 would provide capability to track connection status outside of goka so the above is less critical

- store a `Connected` boolean on the struct implementing goka.Backoff
- set to `false` when Duration() is called
- set to `true` when Reset() is called

mattburman:
[...]

PR #260, opened 2020-07-08
#239 integrate sarama consumer group

frairon:
* integrate sarama consumer group
* remove sarama-cluster dependency


fkarakas:
Why did u remove confluent lib for sarama ? i would like to know the reason plz
thanks @frairon

frairon:
Hi @fkarakas, you mean the support for the C implementation of kafka (I think it was based on librdkafka?), right?

The thing was, that this implementation follows a completely different logic to rebalance and receive messages. In the "old" goka version, we had to do quite some magic to support both sarama and librdkafka, so it was very hard to understand, maintain, and it behaved flaky in an unstable kafka cluster. So we decided to simplify the logic and get rid of it for now.
If this is something that people still need, I think we can somehow get it working again by trying to bridge its behavior to the new sarama logic of the consumer group. But haven't looked into it at all so far.
If you are willing to work on that, PRs are always welcome :D

fkarakas:
Thanks, if i understand correctly it was difficult to make an abstraction in the goka API which can both work with confluent and sarama.

frairon:
Yep exactly, and the point of the refactoring was to simplify things, get rid of sarama-cluster. Are you using the confluent api?

fkarakas:
Yes we are using the confluent lib, as you mentioned it is based on the C librdkafka so it is not a full GO lib which can be a stopper for go purist, but we decided to use it because it is confluent official lib for GO so we think it is "safer" regarding its maintenance/upgrade in the future ...
And in kafka, a big part of the "logic" is implemented in the client side instead of the server side, so it was another +1 in the choice of confluent-kafka-go. This was our conclusion.

frairon:
That sounds like a reasonable choice. So is this a blocker for you to upgrade to a new goka version? And would you have the resources to look into it?

R053NR07:
Hey @fkarakas,
as you said the "pure GO" implementation was the main driver for the decision. CGO always was and is a problem for us, we had to take extra effort to use goka since this was a requirement. Also goka internally we focused mainly on sarama and supporting confluentlib would have been extra effort. We decided to focus on sarama since it is widely used in the go world for working with kafka and we did not see any mayor limitations or drawbacks when using it. 

Regarding the support. Sure librdkafka may get longer support but even if sarama would be deprecated today we would be able to integrate and switch to librdkafka; for us the effort would be almost the same then designing goka with interchangeable kafka libs.

But still. If that's a blocker for you and others we can talk about changing this in future releases.

PR #269, opened 2020-09-06
Configure sarama.Logger to use "[Sarama] " prefix

adw1n:
Sarama sets up its logger with `"[Sarama] "` prefix - check https://godoc.org/github.com/Shopify/sarama#pkg-variables. I suggest we do the same here - this way users are given a hint where the log message came from. Users that do not wish to have this prefix can still use `SetSaramaLogger` to override that behavior.

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -79,7 +79,7 @@ func Default() Logger {
 func Debug(gokaDebug, saramaDebug bool) {
 	defaultLogger.debug = gokaDebug
 	if saramaDebug {
-		SetSaramaLogger(&std{debug: true})
+		SetSaramaLogger(&std{debug: true, prefix: "[Sarama] "})
 	}
 }
 
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -79,7 +79,7 @@ func Default() Logger {
 func Debug(gokaDebug, saramaDebug bool) {
 	defaultLogger.debug = gokaDebug
 	if saramaDebug {
-		SetSaramaLogger(&std{debug: true, prefix: "[Sarama] "})
+		SetSaramaLogger((&std{debug: true}).Prefix("Sarama"))
 	}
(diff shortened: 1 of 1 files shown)


R053NR07 (approved):

frairon (approved):
yeah, looks good. Why not :) Thanks @adw1n for the contribution

logger/logger.go
```diff
@@ -79,7 +79,7 @@ func Default() Logger {
 func Debug(gokaDebug, saramaDebug bool) {
 	defaultLogger.debug = gokaDebug
 	if saramaDebug {
-		SetSaramaLogger(&std{debug: true})
+		SetSaramaLogger(&std{debug: true, prefix: "[Sarama] "})
```
frairon, line 82: @adw1n  Actually I'd like some change, if you don't mind, because this would be inconsistent with how this prefixing approach works.
The `std`-logger field prefix represents the current prefix in brackets e.g. `["Sarama"]`and a prefix path, which allows to extend the prefix, so the prefix path needs to be without brackets. In this case it would need to be `[]string{"Sarama"}`. So when adding a new prefix, e.g. `log.Prefix("Consumer")` we would have a new prefix like `[Sarama>Consumer]`.

If we initialize it like in the PR, the prefix stacking would be broken, and adding a new prefix would produce a prefix like `[[Sarama]>NewPrefix]`. 
This is not very relevant at the moment, because sarama won't call `Prefix` of course, still I think it should be consistent.
To avoid duplication, the `Prefix`-function you could use the Prefix-function after initalization. Like this:
```
SetSaramaLogger((&std{debug: true}).Prefix("Sarama"))
```

What do you say?

adw1n: @frairon I agree. I commited and pushed the requested change.

frairon (approved):

PR #270, opened 2020-09-18
Topic bug in emitterB for the multiple-example fixed.

FredrikBakken:


frairon (approved):
:man_facepalming: thanks

Issue #273, opened 2020-10-09
Processor fails to start with "resource temporarily unavailable" error


mrahs:
This error happens with processors that have `Join` and/or `Persist` edges which use LevelDB. 

The database is opened in a [separate goroutine](https://github.com/lovoo/goka/blob/master/partition_table.go#L203), but the parent one can [return too early](https://github.com/lovoo/goka/blob/master/partition_table.go#L216) without capturing the db reference. In other words, the database is never closed keeping the lock file acquired and causing subsequent attempts to open the database to fail with `resource temporarily unavailable` error. 

In my case, a partition reassignment canceled the session context before the db was fully open.

I will submit a PR to fix this.

adw1n:
@mrahs this got fixed in https://github.com/lovoo/goka/pull/274, right? Can we close this issue?

PR #274, opened 2020-10-09
Processor fails to start with "resource temporarily unavailable" error

mrahs:


mrahs:
Closes #273

partition_table.go
```diff
@@ -194,43 +194,58 @@ func (p *PartitionTable) Close() error {
 func (p *PartitionTable) createStorage(ctx context.Context) (*storageProxy, error) {
 	var (
 		err  error
+		errs = new(multierr.Errors)
 		st   storage.Storage
 		done = make(chan struct{})
 	)
+
 	start := time.Now()
 	ticker := time.NewTicker(5 * time.Minute)
```
ubunatic, line 203: `ticker` declaration and `defer ticker.Stop` should go inside logging goroutine since it only used there.
The logging is then stopped only through the `done` channel.

partition_table.go
```diff
@@ -194,43 +194,58 @@ func (p *PartitionTable) Close() error {
...
 	defer ticker.Stop()
+	defer close(done)
 	go func() {
-		defer close(done)
-		st, err = p.builder(p.topic, p.partition)
-		if err != nil {
-			return
+		for {
+			select {
+			case <-ctx.Done():
```
ubunatic, line 209: Do you need to wait for the context here also?
I guess just stopping the logging via the `done` channel is fine.
The context is then only used outside of the logging goroutine to track the storage creation

ubunatic (commented):
use ticker only logging + close logging via done channel only

partition_table.go
jomaresch: I would also remove the `case <-ctx.Done():` because if the context is done, we will not stop the storage building. So the logs should also not stop.

partition_table.go
```diff
@@ -194,43 +194,58 @@ func (p *PartitionTable) Close() error {
...
+	}
+	err = st.Open()
+	if err != nil {
+		errs.Collect(st.Close())
+		errs.Collect(fmt.Errorf("error opening storage: %v", err))
+		return nil, errs.NilOrError()
+	}
+
+	// close the db if context was cancelled before the builder returned
+	select {
```
frairon, line 231: The reason we implemented it this way was to be able to shutdown the processor while starting, because sometimes the loading could take a long when leveldb does compaction on loading.
But you're right that it will fail on rebalancing before the load is completed. So I guess this solution is correct but might lead to container kills when the container management cannot cleanly shutdown.

partition_table.go
```diff
@@ -194,43 +194,58 @@ func (p *PartitionTable) Close() error {
...
+	case <-ctx.Done():
+		err = st.Close()
+		// do not report an error for only cancelling the context
```
frairon, line 234: the comment is a bit misleading, as it says we're not reporting the error, but actually the errors are added to the collector?

partition_table.go
mrahs: Good point.

partition_table.go
mrahs: Yup, makes sense.

partition_table.go
mrahs: We only report the context error if `st.Close()` returned error. Otherwise, no error is reported even if context had one.
I'll rephrase to: `only collect context error if Close() errored out`

partition_table.go
mrahs: If I understand correctly, you are referring to the case where this function takes long enough to force a container manager to kill the container. In that case, it's explicit that the shutdown was not clean, rather than hiding this failure and causing others.

frairon: yep, that's what I meant. But you've got a point there and we should do it this way.

partition_table.go
frairon: thanks, much easier to understand

frairon (approved):

jomaresch (approved):

PR #276, opened 2020-10-14
Tester: fix concurrent map access

jomaresch:
- protect `Tester. codecs` from concurrent map access

### DIFF
--- a/tester/tester.go
+++ b/tester/tester.go
@@ -41,7 +41,7 @@ type Tester struct {
 	mClients sync.RWMutex
 	clients  map[string]*client
 
-	mCodecs sync.Mutex
+	mCodecs sync.RWMutex
 	codecs  map[string]goka.Codec
 
 	mQueues     sync.Mutex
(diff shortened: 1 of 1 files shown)


frairon (approved):

tester/tester.go
```diff
@@ -41,7 +41,9 @@ type Tester struct {
 	mClients sync.RWMutex
 	clients  map[string]*client
 
-	codecs      map[string]goka.Codec
+	mCodecs sync.RWMutex
+	codecs  map[string]goka.Codec
```
paulhenke, line 45: idiomatic go would call it `codecsMu`

jomaresch: Jop, but in goka all the Mutex are called like this:
```
mClients sync.RWMutex
mQueues     sync.Mutex
...
```
And i don't want to refactor the whole project, that's why i kept the naming schema and called it `mCodecs sync.RWMutex`

PR #279, opened 2020-11-02
fix broken Signal.WaitForState lock, fix Emitter.Finish data races, fix data races in tests

ubunatic:
- fixed all data races in the tests using sync counters
- Mocks: added mutexes where missing
- Signal: fixed broken WaitForState lock
- Emitter: added Emit mutex to avoid racy emits during e.Finish

### DIFF
--- a/emitter.go
+++ b/emitter.go
@@ -19,7 +19,7 @@ type Emitter struct {
 	topic string
 
 	wg   sync.WaitGroup
-	mu   sync.Mutex
+	mu   sync.RWMutex
 	done chan struct{}
 }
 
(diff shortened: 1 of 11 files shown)


partition_table_test.go
```diff
@@ -372,9 +373,9 @@ func TestPT_load(t *testing.T) {
 			consumer               = defaultSaramaAutoConsumerMock(t)
 			topic                  = "some-topic"
 			partition        int32
-			count            int32
+			count                           = new(int64)
```
frairon, line 376: that looks weird, can't w leave it at int64 and call
`atomic.AddInt64(&count, 1)` instead?
I mean this is technically also correct, but it's kind of off-schema.
Someone correct me if this has become standard-style :)

emitter.go
```diff
@@ -70,17 +67,22 @@ func (e *Emitter) EmitWithHeaders(key string, msg interface{}, headers map[strin
...
-		return e.producer.Emit(e.topic, key, data).Then(func(err error) {
-			e.wg.Done()
-		}), nil
-	} else {
-		return e.producer.EmitWithHeaders(e.topic, key, data, headers).Then(func(err error) {
-			e.wg.Done()
-		}), nil
+
+	// protect e.done channel and e.wg WaitGroup together to reject all new emits after calling e.Finish
+	e.mu.Lock()
```
frairon, line 72: I think we should use an RWMutex and do e.mu.RLock() here, otherwise we introduce a synchronization point and goroutines will not be able to Emit messages concurrently

ubunatic: I think tried `&atomic.AddInt64(&count, 1)` at first, which still gave me data races and I though a mutext would be too much. But maybe i missed something

ubunatic:
I think I resolved all issues. Please review :-)

frairon (approved):

PR #280, opened 2020-11-10
Error/Logging improvement

frairon:
* improve error printing: show only user stack trace of panic
* bugfix tester: shutdown when processors panic
* bugfix error reporting: only report error once

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -1,7 +1,74 @@
 package goka
 
+import (
+	"fmt"
+	reflect "reflect"
+	"regexp"
+	"strings"
+
+	"github.com/go-stack/stack"
+)
+
 var (
 	errBuildConsumer = "error creating Kafka consumer: %v"
 	errBuildProducer = "error creating Kafka producer: %v"
 	errApplyOptions  = "error applying options: %v"
 )
+
+var (
+	// this regex matches the package name + some hash info, if we're in gomod but not subpackages
+	// examples which match
+	// * github.com/lovoo/goka/processor.go
+	// * github.com/lovoo/goka@v1.0.0/view.go
+	// * github.com/some-fork/goka/view.go
+	// examples which do not match
+	// * github.com/something/else
+	// * github.com/lovoo/goka/subpackage/file.go
+	// this regex is used to filter out entries from the stack trace that origin
+	// from the root-package of go (but not the subpackages, otherwise we would not see the stack in the example-tests)
+	// reflect.TypeOf(Processor{}).PkgPath() returns (in the main repo) "github.com/lovoo/goka"
+	gokaPackageRegex = regexp.MustCompile(fmt.Sprintf(`%s(?:@[^/]+)?/[^/]+$`, reflect.TypeOf(Processor{}).PkgPath()))
+)
+
+// userStacktrace returns a formatted stack trace only containing the stack trace of the user-code
+// This is mainly used to properly format the error message built after a panic happened in a
+// processor-callback.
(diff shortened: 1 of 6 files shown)


errors.go
```diff
@@ -1,7 +1,60 @@
...
 	errBuildProducer = "error creating Kafka producer: %v"
 	errApplyOptions  = "error applying options: %v"
 )
+
+var (
+	gokaPackageRegex = regexp.MustCompile(`github.com/lovoo/goka[^/]*/[^/]+$`)
+)
+
+// userStacktrace returns a formatted stack trace only containing the stack trace of the user-code
+// This is mainly used to properly format the error message built after a panic happened in one of the
```
175-kavinski, line 22: "one of the..."

175-kavinski (approved):

175-kavinski (approved):

errors.go
```diff
@@ -1,7 +1,61 @@
...
+)
+
 var (
 	errBuildConsumer = "error creating Kafka consumer: %v"
 	errBuildProducer = "error creating Kafka producer: %v"
 	errApplyOptions  = "error applying options: %v"
 )
+
+var (
+	gokaPackageRegex = regexp.MustCompile(`github.com/lovoo/goka[^/]*/[^/]+$`)
```
ubunatic, line 18: What if sb. uses a fork? Is there  a smarter way to get the "current" package name?

partition_processor.go
```diff
@@ -232,7 +236,7 @@ func (pp *PartitionProcessor) Setup(ctx context.Context) error {
 	pp.runnerGroup.Go(func() error {
 		err := pp.run(runnerCtx)
 		if err != nil {
-			pp.log.Printf("Run failed with error: %v", err)
+			pp.log.Debugf("Run failed with error: %v", err)
```
ubunatic, line 239: Does it make sense to have debug logs in a library like goka?

frairon: I was also wondering that.
The logs are mainly to understand what is going on in goka, when it's doing a rebalance etc. 
 Sarama also has internal logs that might give the user an insight why something behaves like it behaves, e.g. a rebalance because of a slow consumption - which is not necessarily an error.
For cases like this logs makes sense I think, but for the one above, probably not because this error should be logged by the client. I didn't want to remove it so I converted it to debug so we usually don't spam the user.

errors.go
frairon: yep, that's exactly the issue. Maybe we really should use reflection like here?
https://golang.org/pkg/reflect/#Type

errors.go
```diff
@@ -1,7 +1,74 @@
...
+	// * github.com/lovoo/goka/processor.go
+	// * github.com/lovoo/goka@v1.0.0/view.go
+	// * github.com/some-fork/goka/view.go
+	// examples which do not match
+	// * github.com/something/else
+	// * github.com/lovoo/goka/subpackage/file.go
+	// this regex is used to filter out entries from the stack trace that origin
+	// from the root-package of go (but not the subpackages, otherwise we would not see the stack in the example-tests)
+	// reflect.TypeOf(Processor{}).PkgPath() returns (in the main repo) "github.com/lovoo/goka"
+	gokaPackageRegex = regexp.MustCompile(fmt.Sprintf(`%s(?:@[^/]+)?/[^/]+$`, reflect.TypeOf(Processor{}).PkgPath()))
```
ubunatic, line 30: Magic!

ubunatic (approved):

Issue #156, opened 2018-10-30
Proposal: preload and hotstandby loading policies


db7:
**Problem**: In processors with large state, scaling the number of processor instances up or down takes long because the group table partitions have to be loaded from Kafka into the local storage. While a processor instance is recovering the partitions, its assigned partitions are not processed, creating sometime a high lag.

**Proposal**: Create different loading policies to mitigate this problem. We could start with 3 policies:

- `ON_DEMAND`: When a processor instance starts, it first rebalances the partitions. The processor instance only loads the partitions assigned to it before starting to process them. This is default policy and how it currently works. 

- `PRELOAD`: When a processor instance starts, it first loads *all* partitions before doing rebalancing. After rebalance, the processor instance does not further load partitions that are not assigned to it. The local storage of processor instances will be lagging for non-assigned partitions as in `ON_DEMAND` policy.

- `HOTSTANDBY`: When a processor instance starts, it first preloads *all* partitions as in `PRELOAD` policy. After rebalance, the processor instance starts processing assigned partitions and *keeps loading* non-assigned partitions.

The `HOTSTANDBY` policy is nice because scaling down the number of processor instances or the crash of instances does not cause any lag because the remainder processor instances can start processing the dangling partitions right away after rebalance. Of course, it drastically increase the network traffic. That is a trade-off.

**Further steps**: In the future, we could introduce preload and hotstandby policies that don't load all partitions, but just a subset. That would require us to play with the rebalancing policy to guarantee that after rebalancing the instances with preloaded partitions will have a high likelihood of being assigned these partitions.

awdng:
I was wondering if I missed something crucial in the presentation at GoDays 2019 or if the on demand scaling would actually be a problem. In scenarios with high peak loads and auto scaling, spinning up new instances that have to read and build their local state first before being ready to serve requests would be a dealbreaker in many scenarios, wouldn't it? Is it advised to use Goka in more Downstream services where processing speed is non-crucial? Should the setup be overscaled by default? I wonder what the best approach is here.

db7:
Good questions, @awdng. Glad you are trying Goka.

> In scenarios with high peak loads and auto scaling, spinning up new instances that have to read and build their local state first before being ready to serve requests would be a dealbreaker in many scenarios

[...]

PR #283, opened 2020-11-18
#156 processor hot standby

frairon:
this PR adds two new options to goka:

* `WithHotStandby` that is used to keep partitions up to date, that are not part of the current assignment
* `WithRecoverAhead` that makes the processor recover all partitions before attempting to join the group.

This was initially proposed in #156 .
This will speed up rebalance, as processors are already up to date and do not need to recover on new partitions.
However, more ressources are be required.

If HotStandby is used, RecoverAhead should also be used. Otherwise adding a new processor to the group will block the partitions that were assigned to the new processor until it has recovered everything.

Neither the Tester nor the MockAutoConsumer could be used for testing, as they assume only one partition. The function is now tested using a "systemtest" that requires a (local) running kafka cluster.

Also add the run mode to web-monitoring (screenshot shows monitoring of `examples/8-monitoring`)
Screenshot from 2021-04-14 12-21-28


options.go
```diff
@@ -176,6 +177,15 @@ func WithHasher(hasher func() hash.Hash32) ProcessorOption {
 	}
 }
 
+// WithHotStandby configures the processor to keep up partitions that it is not assigned when using
+// multiple processors in a group. This will use more resources, but allows to fail over processors in case of crashes
```
db7, line 181: This sounds like failover is impossible without hotstandby. I think what you want to say here is something like "... in a group. WithHotStandby allows fast processor failover since all partitions are hot in other processor instances, but it requires more resources (eg, network and disk)."

frairon: thanks diogo, updated the comment

integrationtest/processor_test.go
```diff
@@ -94,6 +94,84 @@ func TestErrorCallback(t *testing.T) {
 	}
 }
 
+func TestHeaders(t *testing.T) {
```
frairon, line 97: test was just moved from examples-package to the integration tests, because it does not demonstrate how to use the tester-framework but actually tests something.

processor.go
```diff
@@ -57,7 +57,7 @@ type Processor struct {
 
 	state *Signal
 
-	ctx    context.Context
+	done   chan struct{}
```
frairon, line 60: instead of storing the context (which is bad practice), we'll store a channel that is closed when the processor is done. The purpose is very similar, but this is much cleaner.

jomaresch (dismissed):
Nice one

options.go
frairon: fixed

web/templates/monitor/details_processor.go.html
```diff
@@ -2,497 +2,518 @@
...
-          if(lastStats){
-            // diff to previous stats in seconds
-            var timeDiff = (values.now - lastStats.now) / 1000.0;
+              writeRate = (stats.Writes.Count - lastStats.Writes.Count) / timeDiff;
+              writeBytes = (stats.Writes.Bytes - lastStats.Writes.Bytes) / timeDiff;
+            }
+            lastTableStats.set(this, stats);
+
+            return '<td>' + stats.partition.toFixed(0) + '</td>\n' +
+              '<td>' + status + " (" + runMode + ')</td>\n' +
```
frairon, line 188: the file got reformatted, so the changes are rather small. I only added the runMode to the stats for table-partition and joins (see screenshot in PR description)

jomaresch (approved):

PR #284, opened 2020-11-20
Implement DeferCommit to make goka callbacks asynchronously 'finishable'

frairon:


context.go
```diff
@@ -121,6 +121,11 @@ type Context interface {
 	// Context returns the underlying context used to start the processor or a
 	// subcontext.
 	Context() context.Context
+
+	// AsyncDone will delay the commit of the incoming message until the returned message is called.
+	// Failing to call this function will make the processor never commit this message, so it will consume the message again after
+	// the next restart or rebalance.
+	AsyncDone() func(error)
```
ubunatic, line 128: As discussed, we should call it `Defer` or `DeferDone`. In the description describe what it returns and when and how the `doneFunc` should be called (with a small example that shows the preferred usage). I would also warn the user that this will block the processor and that any returned `doneFunc` must be called eventually.

ubunatic (commented):
LGTM, except for the func name and description. The user must really know what he gets and how to use it. Also a test case that shows how the new logic interacts with emitted data, i.e., how the emit counter is affected, would be nice.

jomaresch (approved):
LGTM

ubunatic (approved):

PR #289, opened 2020-12-03
add group getter in context

qneyrat:


context.go
```diff
@@ -264,6 +267,10 @@ func (ctx *cbContext) Offset() int64 {
 	return ctx.msg.Offset
 }
 
+func (ctx *cbContext) Group() Group {
+	return Group(ctx.graph.group)
```
frairon, line 271: total nit picking, but was there a reason not to use the getter?
```suggestion
	return ctx.graph.Group()
```

qneyrat: No it's a mistake +1.

*GroupGraph can be nil ? I add check nil ?

eko (approved):

context.go
frairon: well technically it cannot be nil, because the context only lives in a processor, which needs a graph to function. So I'd just skip it :)

frairon (approved):

PR #291, opened 2020-12-14
LVPD-843 callback error leads to processor shutdown

frairon:


tester/consumergroup.go
```diff
@@ -83,7 +83,11 @@ func (cg *consumerGroup) Consume(ctx context.Context, topics []string, handler s
 				return nil
 			})
 			errg.Go(func() error {
-				return handler.ConsumeClaim(session, claim)
+				err := handler.ConsumeClaim(session, claim)
```
frairon, line 86: this makes the tester implementation of the consumergroup behave like sarama, i.e. it pushes errors from the consumeclaim into the error channel instead of returning it - similar to sarama.

mmreza79 (approved):
LGTM

ubunatic (approved):
LGTM

PR #293, opened 2020-12-23
fix topic manager stream retention not in milliseconds

tlinford:
Noticed that topics created with TopicManager.EnsureStreamExists had much longer retention than expected, and setting it to -1 to avoid deletion doesn't work either.
%d on a duration seems to print out number of microseconds, and not milliseconds:
```
t := 1 * time.Hour
fmt.Printf("%d", t)
```
gives:
`3600000000000`

### DIFF
--- a/topic_manager.go
+++ b/topic_manager.go
@@ -177,7 +177,7 @@ func (m *topicManager) EnsureStreamExists(topic string, npar int) error {
 		npar,
 		m.topicManagerConfig.Stream.Replication,
 		map[string]string{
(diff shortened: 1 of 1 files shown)


frairon (approved):

frairon:
The whole functionality of creating and maintaining topics required for processors is somehow buggy and obviously needs some love, we noticed that some time ago. Especially when your cluster is configured to auto-create topics, you might end up having topics with the wrong configuration. But we're working on that.
Nevertheless this retention-config was a bug and needed to be fixed, so thanks a lot!

PR #295, opened 2021-01-13
Use Go format for examples in README

heppu:


R053NR07 (approved):

R053NR07:
Seems like we missed that. Thanks :)

PR #297, opened 2021-01-21
topic-manager improvements

frairon:
This PR fixes two main issues:
* when checking if a topic exists and the cluster was configured with `auto.create.topics.enable=true`, the topic would be auto-created with default-configuration. This messes up table-creation because those topics would not be log-compacted which is required by goka. The new version does not trigger auto-creation when checking for a topic.
* if the topic configuration does not match, the topic manager can be configured to 
  * print a warning
  * fail
  * ignore (that was the default before)
  The topic manager does not attempt to modify misconfigured topics, as this could be dangerous for intentional configuration changes via other kafka tools outside of goka.


integrationtest/topicmanager_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	tm, err := goka.TopicManagerBuilderWithConfig(cfg, goka.NewTopicManagerConfig())([]string{"localhost:9092"})
+	test.AssertNil(t, err)
+
+	err = tm.EnsureTopicExists("test10", 4, 2, nil)
+	test.AssertNil(t, err)
+
+}
+
+// Testthe topic manager with sarama version v11 --> so it will test topic configuration using
```
mmreza79, line 36: `Test the`

integrationtest/topicmanager_test.go
```diff
@@ -0,0 +1,107 @@
...
+	tm, err := goka.TopicManagerBuilderWithConfig(cfg, tmc)([]string{"localhost:9092"})
+	test.AssertNil(t, err)
+
+	client, _ := sarama.NewClient([]string{"localhost:9092"}, cfg)
+	admin, _ := sarama.NewClusterAdminFromClient(client)
+
+	t.Run("ensure-new-stream", func(t *testing.T) {
+		var (
+			topic = newTopicName()
+		)
```
mmreza79, line 58: delete parenthesis

topic_manager.go
```diff
@@ -31,14 +31,18 @@ type TopicManager interface {
...
+	admin              sarama.ClusterAdmin
 	client             sarama.Client
 	topicManagerConfig *TopicManagerConfig
 }
 
 // NewTopicManager creates a new topic manager using the sarama library
 func NewTopicManager(brokers []string, saramaConfig *sarama.Config, topicManagerConfig *TopicManagerConfig) (TopicManager, error) {
+
+	if !saramaConfig.Version.IsAtLeast(sarama.V0_10_0_0) {
+		return nil, fmt.Errorf("goka's topic manager needs kafka version v0.10.0.0 or higher to function. Version was %s", saramaConfig.Version.String())
```
mmreza79, line 43: Version was -> Version is

topic_manager.go
```diff
@@ -105,79 +113,175 @@ func (m *topicManager) Close() error {
...
+	// already exists, no need to do create
+	if hasTopic {
 		return nil
 	}
+
 	return m.createTopic(topic,
 		npar,
 		rfactor,
 		config)
 }
```
jomaresch, line 233: can we reduce the nested `if`s and return early, like this?:
```
if !hasTopic {
    return createTopic
}

if !m.adminSupported() {
    return nil
}

...
configMismatch code
...

```

jomaresch: or do i miss something?

frairon: nope, you're right, it was kind of messy. Tried to improve it

jomaresch (dismissed):
LGTM

graph_test.go
```diff
@@ -77,6 +77,14 @@ func TestGroupGraph_Validate(t *testing.T) {
 
 }
 
+func TestGroupGraph_chainEdges(t *testing.T) {
+	test.AssertEqual(t, len(chainEdges()), 0)
+	test.AssertEqual(t, len(chainEdges(Edges{}, Edges{})), 0)
+	test.AssertEqual(t, chainEdges(Edges{Join("a", nil)}, Edges{}), Edges{Join("a", nil)})
+	test.AssertEqual(t, chainEdges(Edges{Join("a", nil)}, Edges{Join("a", nil), Join("b", nil)}), Edges{Join("a", nil), Join("a", nil), Join("b", nil)})
+
```
R053NR07, line 85: :grimacing:  ^^

topic_manager.go
```diff
@@ -105,70 +113,161 @@ func (m *topicManager) Close() error {
...
+	}
 
-	request := sarama.CreateTopicsRequest{
-		Timeout:      time.Second * 15,
-		TopicDetails: topicDetails,
+	return nil
+}
+
+func (m *topicManager) handleConfigMismatch(message string) error {
+	switch m.topicManagerConfig.MismatchBehavior {
```
R053NR07, line 162: Do you think this switch is the best option?
If you want a default case I would rather add it explicitly as long every other case also results in a return of nil or an error. Otherwise I would not return in the first case since it's basically just an addition to the default case and thus a redundant code path? But since it's currently equal it does not make a difference.

R053NR07 (dismissed):
Think this looks good.

topic_manager.go
```diff
@@ -31,14 +31,18 @@ type TopicManager interface {
...
+	admin              sarama.ClusterAdmin
 	client             sarama.Client
 	topicManagerConfig *TopicManagerConfig
 }
 
 // NewTopicManager creates a new topic manager using the sarama library
 func NewTopicManager(brokers []string, saramaConfig *sarama.Config, topicManagerConfig *TopicManagerConfig) (TopicManager, error) {
+
+	if !saramaConfig.Version.IsAtLeast(sarama.V0_10_0_0) {
+		return nil, fmt.Errorf("goka's topic manager needs kafka version v0.10.0.0 or higher to function. Version is %s", saramaConfig.Version.String())
```
frairon, line 43: Question: do we need v0.10.0.0 now? is this a breaking change?

topic_manager.go
```diff
@@ -105,70 +113,161 @@ func (m *topicManager) Close() error {
...
 func (m *topicManager) createTopic(topic string, npar, rfactor int, config map[string]string) error {
+	m.topicManagerConfig.Logger.Debugf("creating topic %s with npar=%d, rfactor=%d, config=%#v", topic, npar, rfactor, config)
 	topicDetail := &sarama.TopicDetail{}
 	topicDetail.NumPartitions = int32(npar)
 	topicDetail.ReplicationFactor = int16(rfactor)
 	topicDetail.ConfigEntries = make(map[string]*string)
 
 	for k, v := range config {
-		topicDetail.ConfigEntries[k] = &v
+		// need to make a copy, or the config values will contain only one value
```
frairon, line 147: Question: what does that mean?

topic_manager.go
frairon: Goka requires kafka version `v0.10.2.0` since version v1.0.0, as the consumer-group interface requires this. I guess we could check that in advance, but sarama will fail after creating the first processor anyway.

topic_manager.go
frairon: added a comment

topic_manager.go
frairon: pushed another version. Take a look if that's better. Not entirely sure what you meant here :)

R053NR07 (approved):
LGTM

vladlepbeat:
Hi @frairon, 
We updated to goka 1.0.6 recently and noticed this is a breaking change for us when spinning up testing environments. To explain a bit the use case, we use goka for a data pipeline and all the data we collect is deleted after 15 minutes. Although in goka compaction should be used, in our case it is actually not needed as everything is removed in 15 minutes. In those 15 mins compaction would not kick in anyways. We have testing environments that we can spin up with their own kafka cluster when needed. For these we have `auto.create.topics.enable=true` and in the past topics where created and all was working ok for us. 

To fix the issue, we now downgraded to an older version but wanted to ask if we could have this configurable?

jomaresch:
@vladlepbeat 
As far as I know you can call [EnsureStreamExists()](https://github.com/lovoo/goka/blob/eaaa070c7b9320aa293042028067802d484670b5/topic_manager.go#L17) on every startup for each topic you need.

Here is an example:
https://github.com/lovoo/goka/blob/0085fa7f7c2e542e740ec565af7662b2d441b019/examples/1-simplest/main.go#L115-L118

jomaresch:
Nice.
Your are welcome 🙂

frairon:
@vladlepbeat the thing with automatic topic creation was, that it was actually a bug in goka. We used the wrong API call so to say, which made clusters configured with auto-create create the topics also for tables. Then we ended up with table topics configured as streams (no compaction), which completely messes up the state in the views.
There was a similar discussion going on in issue #323 (which isn't solved yet).
If remembering correctly the bottom line is, that we want to implement a hybrid approach that makes goka manage the topics it owns (like tables, loops and output topics) and let the cluster decide for the other topics. So there's still a bit of work needed here. But Johannes' suggestion is the perfect workaround so far.

PR #298, opened 2021-01-25
Allow context to set Kafka headers

AlecY:
Added additional context functions which support setting Kafka headers

### DIFF
--- a/context.go
+++ b/context.go
@@ -11,7 +11,7 @@ import (
 	"github.com/lovoo/goka/multierr"
 )
 
-type emitter func(topic string, key string, value []byte) *Promise
+type emitter func(topic string, key string, value []byte, headers map[string][]byte) *Promise
 
 // Context provides access to the processor's table and emit capabilities to
 // arbitrary topics in kafka.
@@ -71,7 +71,7 @@ type Context interface {
 	// This method might panic to initiate an immediate shutdown of the processor
 	// to maintain data integrity. Do not recover from that panic or
 	// the processor might deadlock.
-	SetValue(value interface{})
+	SetValue(value interface{}, options ...ContextOption)
 
 	// Delete deletes a value from the group table. IMPORTANT: this deletes the
 	// value associated with the key from both the local cache and the persisted
@@ -80,7 +80,7 @@ type Context interface {
 	// This method might panic to initiate an immediate shutdown of the processor
 	// to maintain data integrity. Do not recover from that panic or
 	// the processor might deadlock.
-	Delete()
+	Delete(options ...ContextOption)
 
 	// Timestamp returns the timestamp of the input message. If the timestamp is
 	// invalid, a zero time will be returned.
@@ -105,7 +105,7 @@ type Context interface {
 	// This method might panic to initiate an immediate shutdown of the processor
 	// to maintain data integrity. Do not recover from that panic or
 	// the processor might deadlock.
-	Emit(topic Stream, key string, value interface{})
+	Emit(topic Stream, key string, value interface{}, options ...ContextOption)
 
 	// Loopback asynchronously sends a message to another key of the group
 	// table. Value passed to loopback is encoded via the codec given in the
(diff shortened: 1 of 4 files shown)


frairon:
Hey @AlecY, 
many thanks for the contribution! The code looks good in general, but there's one issue: the headers are sent to Kafka but they're not persisted to local disk, so neither views nor the processor itself can access them, which seems inconsistent.
So I'm wondering what would be the use case behind having this functionality?
Thanks a lot!

AlecY:
Hey @frairon,
My use case for this functionality is using Kafka headers to send tracing information between services. I have services that read another services persist topic as an input for further processing and I'd like tracing information to be written into the persist topic so it can be read when it's processed as an input. I agree that it would be more consistent if headers where persisted to disk. However I don't require it for my use case, and I don't see a way to add header information to storage without a breaking change.

frairon:
Hey AlecY, 
your use case makes sense and we should definitely extend the interface to support it one way or another.

We had some internal discussions how to approach this, because to be completely honest, I'm
not too happy with adding an extra version of each function just to support headers. 

We thought of two alternatives:
We could store the headers in the context prior to usage, like this:

```go
func callback(ctx goka.Context, msg interface{}){

  // setting the header, so the context will use them on every call
  // where it can be applied.
  ctx.SetHeaders(map[string]{
    "key": []byte{"value"},
    })

  // setvalue/emit with headers
  ctx.SetValue(..)
  ctx.Emit(..)

  // suppose we want some calls not to send the headers anymore
  ctx.ClearHeaders()
  // emit without headers
  ctx.Emit(..)
}

```
This approach especially makes sense if the headers are identical within one callback execution.


One other alternative (we'd prefer that :)), would be to add extendable options to the context-functions.
This means no api-change and a great deal of flexibility if there are more options to be added in future.
Something like this:

```go

type ctxOptions struct{
  emitHeaders map[string][]byte
}

type ContextOption func(*ctxOptions)

// CO-prefix to make it distinguishable from the Processor/Emitter/View-Options, but there
// are probably better names
func WithCOEmitHeaders(headers map[string][]byte) ContextOption{
  return func(opts *ctxOptions){
    opts.headers = headers
  }
}

// the context-interface would be modified like this:
type Context interface {
  // ...	
  Emit(topic Stream, key string, value interface{}, options ...ContextOption)
    
  Loopback(key string, value interface{}, options ...ContextOption)
  // modify the other functions too...
}


// how to use it in a callback
func callback(ctx context.Context, msg interface{}){
  // emit with headers
  ctx.Emit("topic", "key", "some-message", WithCOEmitHeaders(map[string][]byte{"header":[]byte{"value"}}))

  // emit without headers
  ctx.Emit("topic", "key", "some-message")
}

```

But again, those are just rough proposals, maybe there are even more elegant solutions, so we're open for discussions how we could solve this issue.

AlecY:
Hi @frairon,

Thank you for taking the time to post such a detailed reply. I hope I've understood your comment correctly that you prefer the second alternative you posted. I have updated the pull request to use an option based approach. If I have misunderstood you please let me know and I can switch to a solution based on your first suggestion.

context.go
```diff
@@ -170,8 +170,16 @@ type cbContext struct {
 	wg     *sync.WaitGroup
 }
 
+func WithCtxEmitHeaders(headers map[string][]byte) ContextOption{
```
frairon, line 173: A small description would be nice :)

options.go
```diff
@@ -519,3 +519,15 @@ func (opt *eoptions) applyOptions(topic Stream, codec Codec, opts ...EmitterOpti
 		opt.builders.topicmgr = DefaultTopicManagerBuilder
 	}
 }
+
+type ctxOptions struct{
+	emitHeaders map[string][]byte
+}
+
+type ContextOption func(*ctxOptions)
```
frairon, line 527: little comment would be nice here too.

frairon (commented):
Awesome, that was exactly what I meant. Just didn't want to "dictate" the preferred solution, in case you had other ideas. Just would like the two missing comments so the go linter stops complaining and it's good to go

options.go
AlecY: I've added a comment now.

context.go
AlecY: Oops, I've added a comment. I also noticed this should probably be in the options file and moved it.

frairon (approved):
looks good to me. Any objections to merge, @R053NR07?

R053NR07 (approved):

R053NR07:
No, this is awesome :)

AlecY:
Excellent, thank you for being so responsive and helpful!

PR #303, opened 2021-03-02
Improve goka.Context docs

adw1n:
Please squash on merge.

### DIFF
--- a/context.go
+++ b/context.go
@@ -123,14 +123,14 @@ type Context interface {
 
 	// Context returns the underlying context used to start the processor or a
 	// subcontext.
+	// *Important*: this context is only safe to use within this goroutine, so do not pass it to an
+	// asynchronously called callback.
 	Context() context.Context
 
 	// DeferCommit makes the callback omit the final commit when the callback returns.
 	// It returns a function that *must* be called eventually to mark the message processing as finished.
 	// If the function is not called, the processor might reprocess the message in future.
 	// Note when calling DeferCommit multiple times, all returned functions must be called.
-	// *Important*: this context is only safe to use within this goroutine, so do not pass it to an
-	// asynchronously called callback.
 	DeferCommit() func(error)
 }
 
--- a/context.go
+++ b/context.go
@@ -122,15 +122,15 @@ type Context interface {
 	Fail(err error)
 
 	// Context returns the underlying context used to start the processor or a
-	// subcontext.
-	// *Important*: this context is only safe to use within this goroutine, so do not pass it to an
-	// asynchronously called callback.
+	// subcontext. Returned context.Context can safely be passed to asynchronous code and goroutines.
 	Context() context.Context
 
 	// DeferCommit makes the callback omit the final commit when the callback returns.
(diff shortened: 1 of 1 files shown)


context.go
```diff
@@ -123,14 +123,14 @@ type Context interface {
 
 	// Context returns the underlying context used to start the processor or a
 	// subcontext.
+	// *Important*: this context is only safe to use within this goroutine, so do not pass it to an
```
frairon, line 126: Just checked the commit that added `DeferCommit`, and it was not moved and actually intended to be used at `DeferCommit`. Just the wording is a bit confusing. 

Calling `DeferCommit` usually means you need some asynchronous call, otherwise you wouldn't need `DeferCommit`. 
What this comment intended to mean: Do not pass the `goka.Context` into that asynchronous call, because it's not valid to be used anymore.
The `context.Context` that is available by `ctx.Context()` however, can safely be passed to asynchronous code, and should be, to make it stop when the processor shuts down, it's a child-context of the processor's context.

For example, the processor needs to call a grpc function, but the processor should not be blocked waiting for the call. 
(Regardless if that's a good practice or not :)...)
```go

// GOOD:
func handleMsg(ctx goka.Context, msg interface{}){
    // ...
    commit := ctx.DeferCommit()
    callbackCtx := ctx.Context()
    go func(){
        err := grpcClient.Foo(callbackCtx, ...)
        commit(err)
    }()
}
```
So here we're passing the `context.Context` of the `goka.Context` into the goroutine, so the grpc-client can abort the call when the processor shuts down, so the context will appropriately finished (successfully or not). We do not use `goka.Context` inside the goroutine.


How it should be NEVER done:
```go
// BAD:
func handleMsg(ctx goka.Context, msg interface{}){
    // ...
    commit := ctx.DeferCommit()
    go func(){
        // Emit/Value/SetValue/Join/Lookup etc. lead to inconsistent results and break the general contracts of goka
        ctx.Emit(...)
        commit(...)
    }()
}
```
That's what the comment was supposed to mean.
So maybe leave the comment at `DeferCommit`, would the following wording be a bit more clear? 
```go
// *Important*: the context where `DeferCommit` is called, is only safe to use within this callback, 
// never pass it into asynchronous code or goroutines
...
```
Feel free to change or come up with something new.

adw1n: > 	// *Important*: this context is only safe to use within this goroutine, so do not pass it to an
	// asynchronously called callback.
	
This comment left me scratching my head for the following reasons:
* "this context" - what context are they referring to? Was it a typo and they meant to say callable?
* If the context is not safe to be passed is the callable also not safe to be passed?
* Why are they talking about the context here? Wouldn't `goka.Context`/`goka.Context::Context` be a better place to document that?


> The context.Context that is available by ctx.Context() however, can safely be passed to asynchronous code, and should be, to make it stop when the processor shuts down, it's a child-context of the processor's context.

This is great! Let's document that.

> would the following wording be a bit more clear?

Much better although it repeats:
> // The context is only valid within the callback, do not store it or pass it to other goroutines.

frairon (approved):
sounds good to me

Issue #307, opened 2021-03-12
Support Kafka headers in tester package


wwaldner-amtelco:
Recently PR #77 was added that allows passing headers to the context emit method.  The same needs to be done for the Tester.Consume(...) method.  In order to write unit tests that can properly test processors that utilize headers in the handling of a message, the tester interface needs to accept and pass headers to processors.

wwaldner-amtelco:
I have created a PR #308 that implements and tests this feature.

frairon:
should be closed, PR is merged. Feel free to reopen/create new issue if there are more problems.

PR #308, opened 2021-03-13
Initial support for headers in tester

wwaldner-amtelco:
This is a PR for issue #307.  I would appreciate any feedback.

### DIFF
--- a/tester/consumer.go
+++ b/tester/consumer.go
@@ -107,16 +107,8 @@ type partConsumerMock struct {
 func (pcm *partConsumerMock) catchup() int {
 	var numCatchup int
 	for _, msg := range pcm.queue.messagesFromOffset(pcm.hwm) {
-		headers := make([]*sarama.RecordHeader, len(msg.headers))
-		idx := 0
-		for k, v := range msg.headers {
-			headers[idx] = &sarama.RecordHeader{
-				Key:   []byte(k),
-				Value: v,
-			}
-		}
 		pcm.messages <- &sarama.ConsumerMessage{
-			Headers:   msg.SaramaHeaders(),
+			Headers:   msg.saramaHeaders(),
 			Key:       []byte(msg.key),
 			Value:     msg.value,
 			Topic:     pcm.queue.topic,
--- a/tester/consumer.go
+++ b/tester/consumer.go
@@ -107,6 +107,14 @@ type partConsumerMock struct {
 func (pcm *partConsumerMock) catchup() int {
 	var numCatchup int
 	for _, msg := range pcm.queue.messagesFromOffset(pcm.hwm) {
+		headers := make([]*sarama.RecordHeader, len(msg.headers))
+		idx := 0
+		for k, v := range msg.headers {
+			headers[idx] = &sarama.RecordHeader{
+				Key:   []byte(k),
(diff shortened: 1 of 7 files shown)


tester/queue.go
```diff
@@ -2,12 +2,27 @@ package tester
...
-	value  []byte
+	offset  int64
+	key     string
+	value   []byte
+	headers map[string][]byte
+}
+
+func (m *message) SaramaHeaders() []*sarama.RecordHeader {
+	headers := make([]*sarama.RecordHeader, len(m.headers))
+	idx := 0
```
kavinsk1, line 18: will be always 0, which is not intended I guess

tester/consumer.go
```diff
@@ -107,7 +107,16 @@ type partConsumerMock struct {
 func (pcm *partConsumerMock) catchup() int {
 	var numCatchup int
 	for _, msg := range pcm.queue.messagesFromOffset(pcm.hwm) {
+		headers := make([]*sarama.RecordHeader, len(msg.headers))
```
juandes, line 110: Hi :)

To clarify, where is the `headers` list been used? I just see you are assigning `RecordHeader`'s to it but I don't see what's using `headers`. 

Also, this is similar to what you have in the `SaramaHeaders()` function from tester/queue.go (line 16).

Thanks!

tester/queue.go
```diff
@@ -2,12 +2,27 @@ package tester
...
-	offset int64
-	key    string
-	value  []byte
+	offset  int64
+	key     string
+	value   []byte
+	headers map[string][]byte
+}
+
+func (m *message) SaramaHeaders() []*sarama.RecordHeader {
```
juandes, line 16: Since it is an exported function, could we add a small comment explaining it?

tester/queue.go
wwaldner-amtelco: New commit takes care of that.  If forgot to remove unused code that I had factored into a method on the tester.message struct

tester/consumer.go
wwaldner-amtelco: Removed from new commit

tester/queue.go
wwaldner-amtelco: now private and commented

tester/queue.go
frairon: I guess this comment was meant for the other code comment, where it was actually removed. Here, the variable `idx` is still always zero. Rewriting the unit test to contain multiple header values would fail, because only the first is not-nil.

I guess it should be like:
```go
headers := make([]*sarama.RecordHeader, 0, len(m.headers))
for k, v := range m.headers {
headers = append(headers, &sarama.RecordHeader{
	Key:   []byte(k),
	Value: v,
})
}
```

wwaldner-amtelco: Yes, that would work.  I meant to increment idx in the loop.  I will correct this.

frairon: yes of course, just proposing an alternative to get rid of the counter variable

wwaldner-amtelco: I agree, much cleaner.  Committed test and fix.

juandes (approved):

juandes:
Thanks for the changes :)

PR #309, opened 2021-03-14
fixed tests

frairon:


mmreza79 (approved):

mmreza79:
One question. Is there any difference between `Emit` and `EmitWithHeaders` in this PR? I don't know why did you changed it even you are not passing any header info.

frairon:
For this test there is no difference, but the underlying context-implementation changed. A producer has the methods `Emit` and `EmitWithHeaders`, those are used when e.g. you call `ctx.Emit(..)`, because the producer-method is passed to the callback-context here: https://github.com/lovoo/goka/blob/master/partition_processor.go#L497
So whatever is passed there, will be called (whether headers are used or not). 
And since [this](https://github.com/lovoo/goka/commit/373b87deda5fc209d80cddfe9a3dc559d0d0b1e4#diff-096176ac41b335d614f611076ff3d39112ff16596f0c2fee794b29a6e10ec254R497) commit, `EmitWithHeaders` is always used, so the mock has to expect that too or the tests will fail.

Issue #312, opened 2021-03-25
Headers option should append headres not replace them


mrahs:
`WithCtxEmitHeaders` option [currently replaces headers](https://github.com/lovoo/goka/blob/68893ef0b6d9748514f3593445498b592ae3911a/options.go#L534) every time it's called, which means only the last call gets its headers registered. I would expect multiple uses of this option to merge headers instead.

So roughly this:
```
func WithCtxEmitHeaders(headers map[string][]byte) ContextOption{
	return func(opts *ctxOptions){
                for k, v := range headers { opts.emitHeaders[k] = v }
	}
}
```
Instead of this:
```
func WithCtxEmitHeaders(headers map[string][]byte) ContextOption{
	return func(opts *ctxOptions){
		opts.emitHeaders = headers
	}
}
```

frairon:
good point @mrahs, will [this](https://github.com/lovoo/goka/pull/313/files) fix the issue for you?

mrahs:
Yep. Thanks @frairon for the quick fix!

frairon:
Awesome, thanks for the hint. PR is merged, will close this issue.

PR #313, opened 2021-03-26
issue #312 append headers on emit instead of replacing them

frairon:


examples/4-tests/example_test.go
```diff
@@ -353,10 +353,16 @@ func Test_7InputOutputWithHeaders(t *testing.T) {
 	proc, _ := goka.NewProcessor([]string{}, goka.DefineGroup("group",
 		goka.Input("input", new(codec.String), func(ctx goka.Context, msg interface{}) {
 			processorHeaders = ctx.Headers()
-			ctx.Emit("output", ctx.Key(), fmt.Sprintf("forwarded: %v", msg), goka.WithCtxEmitHeaders(
-				map[string][]byte{
-					"Header1": []byte("to output"),
-				}))
+			ctx.Emit("output", ctx.Key(), fmt.Sprintf("forwarded: %v", msg),
+				goka.WithCtxEmitHeaders(
+					map[string][]byte{
```
mrahs, line 358: Would be great to add colliding headers as well to assert overwrite behaviour, but not critical.

frairon: No problem, just added it.

examples/4-tests/example_test.go
```diff
@@ -353,10 +353,18 @@ func Test_7InputOutputWithHeaders(t *testing.T) {
...
+						"Header1": []byte("to output"),
+						"Header2": []byte("to output2"),
+					}),
+				goka.WithCtxEmitHeaders(
+					map[string][]byte{
+						"Header2": []byte("to output3"),
+						"Header3": []byte("to output4"),
```
mmreza79, line 365: Is it header `Header2, Header3`, or `Header3 Header4`?

examples/4-tests/example_test.go
```diff
@@ -376,6 +384,11 @@ func Test_7InputOutputWithHeaders(t *testing.T) {
 			"Header1": []byte("value 1"),
 			"Header2": []byte("value 2"),
 		}),
+		tester.WithHeaders(
+			map[string][]byte{
+				"Header2": []byte("value 3"),
+				"Header3": []byte("value 4"),
```
mmreza79, line 390: same question

examples/4-tests/example_test.go
frairon: ok I admit the values are a bit misleading, but the idea here was to test the overwriting of header values. `Header2` is in both header-options and gets overwritten by the "last version".

examples/4-tests/example_test.go
frairon: same answer :)

mmreza79 (approved):

PR #315, opened 2021-04-23
Improve headers implementation

mrahs:
This patch adds default headers option to the emitter and the processor. It helps when some headers are defined per emitter/processor. Without it, developers are forced to track every (new) emit call or create custom wrappers.

I've also updated the callback for partition tables to pass on message headers.
I used sarama's type for the table-update callback to leverage varargs for backwards compatibility.

I believe headers should now be covered everywhere. At least everywhere I've been to!


headers.go
```diff
@@ -0,0 +1,40 @@
+package goka
```
mrahs, line 1: If breaking changes are accepted, I'd like to create `type Headers map[string][]byte` and define the functions below with that type as the receiver. It would make the code base more readable and future enhancements more flexible.

Otherwise, maybe just an alias for the readability part: `type Headers = map[string][]byte` which will keep everything backwards compatible and the change would be just a rename.

mrahs:
@frairon I would appreciate your feedback on this PR.

frairon:
sorry @mrahs, busy week. I'll try to give it a proper look in the evening (I mean evening in europe :D)

headers.go
frairon: To be honest I don't mind the breaking change, especially since it will be obvious after an upgrade and can be fixed with simple replacement.
What's your opinion on that @jomaresch @R053NR07  @mmreza79?

frairon (approved):
looks good to me so far. If there are no objections about introducing the `Headers` type from others, feel free to change that

headers.go
jomaresch: I agree. ✅
Having `type Headers map[string][]byte` will improve the readability. 
I also don't mind breaking changes.

jomaresch (dismissed):

headers.go
frairon: awesome, @mrahs let's do it

juandes (approved):

headers.go
mrahs: Thanks. I'll submit another PR for that change.

mrahs: Ah.. looks like I ended up updating this PR.

frairon (approved):
looks good to me, many  thanks @mrahs for the contribution!!

PR #317, opened 2021-04-24
NewTopicManagerConfig difference between comment and code

breskos:
Hi LOVOOians, cleared up a difference between comment and code in NewTopicManagerConfig.

### DIFF
--- a/topic_manager.go
+++ b/topic_manager.go
@@ -358,8 +358,8 @@ func (tmc *TopicManagerConfig) tableCleanupPolicy() string {
 // initialize the config with reasonable defaults
 func NewTopicManagerConfig() *TopicManagerConfig {
 	cfg := new(TopicManagerConfig)
-	cfg.Table.Replication = 2
-	cfg.Stream.Replication = 2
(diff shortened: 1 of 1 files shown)


topic_manager.go
```diff
@@ -358,8 +358,8 @@ func (tmc *TopicManagerConfig) tableCleanupPolicy() string {
 // initialize the config with reasonable defaults
 func NewTopicManagerConfig() *TopicManagerConfig {
 	cfg := new(TopicManagerConfig)
-	cfg.Table.Replication = 2
-	cfg.Stream.Replication = 2
+	cfg.Table.Replication = 1
```
frairon, line 361: Actually I'd say replication factor 2 by default makes more sense, so why not update the comment instead? Also since this is has been the default configuration for a longer time, wouldn't it be kind of a breaking change and force everyone to update their config just to make sure to continue using replication factor of 2?

breskos: make sense.

frairon:
closing this, we updated the comment.

Issue #318, opened 2021-05-05
LevelDB's Seek semantics do not honour goka's


mrahs:
Goka's interface requires clients to call `Next` after a successful `Seek` before accessing the key-value pair:
- https://github.com/lovoo/goka/blob/48c2ec23705ff6ea23bb760e832cf2705d559edc/iterator.go#L24
- https://github.com/lovoo/goka/blob/48c2ec23705ff6ea23bb760e832cf2705d559edc/storage/storage.go#L36

LevelDB's interface makes the key-value pair immediately available after a successful `Seek`:
- https://github.com/syndtr/goleveldb/blob/64b5b1c739545ed311fb9d9924d19d188fabdc83/leveldb/iterator/iter.go#L39

Calling `Next` on a leveldb implementation after a successful `Seek` effectively skips an element in the iterator.

I would expect all implementations of the interface to behave the same.

Suggested fix is adding another edge-case to goka's wrapper to skip the first `Next` call after a `Seek`: https://github.com/lovoo/goka/blob/48c2ec23705ff6ea23bb760e832cf2705d559edc/storage/iterator.go#L15

frairon:
Hi Anas, 
awesome observation, I think the Seek was used so rarely that it never occured.
Let's add the edge case. 

Will you give it a shot or should I?

Cheers!

frairon:
the fix is merged, so I'll close the issue.

Issue #181, opened 2019-04-05
Seek in memory.go appears to be broken


abhishekrb19:
`Seek()` in `memory.go` appears to be broken. It would always return `true`, regardless of the key being seeked for. 
```golang
func (i *memiter) Seek(key []byte) bool {
	seek := make(map[string][]byte)
	keys := []string{}
	for k, v := range i.storage {
		if strings.Contains(k, string(key)) {
			keys = append(keys, k)
			seek[k] = v
		}
	}
	i.current = -1
	i.storage = seek
	i.keys = keys
	return !i.exhausted()
}
```

`i.exhausted()` will always return `false` from `Seek()` because current is being overwritten to -1. Also, should the check for keys actually be `strings.HasPrefix(k, string(key))` and not `strings.Contains(k, string(key))`?

frairon:
Hi @abhishekrb19, 
thanks for reporting the issue!
I'll have a look on that. Can't promise anything, but I'll try to do it soonish. 
As usual, improvements and PRs are very welcome :) 

Cheers

db7:
Actually, there is an unmerged branch by Sami that should have fixed these issues. May make sense to check that out.

db7:
@frairon can you check the PR by Sami? It fixes this issue.

frairon:
seems fixed.

PR #320, opened 2021-05-07
#318 avoid losing an element in leveldb seeker

mrahs:
This fix covers iterators for partition tables.
Views use the merge iterator which was not affected by the bug due to calling `Next` upon initialization and caching the pair before calling `Next` again (which was perfect for `Seek`).

That means the merge iterator would have paniced with other implementations such as in-memory which is used the tester package. I have added a test case to demonstrate the panic and the fix.


frairon (commented):
The code looks good, it should fix the issue. Also thanks for the regression test!

But I'm actually wondering whether we should just remove that comment? Checking the history I found that the comment was added by a bigger refactoring last week, but the iterator implementation actually did not change as far as I remember.
So maybe just the comment is wrong and after removing it we should be fine?
What do you think?

mrahs:
It's not just a comment. There are two implementations with different semantics. In this patch, I changed the leveldb one to match the comment on the common interface. I'm fine if patching the memory iterator is preferred. The important thing is that all implementations have the same semantics.

mrahs:
Just noticed there is an old related issue: https://github.com/lovoo/goka/issues/181

frairon:
Yep, so it seems that something is quite odd here. I took the comments from the old branch mentioned in #181 because I thought it was improving storage, but actually that branch was adding some layer to track offsets along with keys for some retention-cleanup that we will probably not add.
So I'd say, the comment is wrong, the storage doesn't need a call to `Next` after `Seek`. 
The memory-storage is a different thing though. It's mostly used for testing, the Seek-Iterate is semantically different from the other storages, because it's not ordered. When seeking it's just selecting the keys that contain the seeked key somehow, which is most likely not what users would expect from the storage. However, as said it's just used for testing (unless you know what you're doing), so we should probably just make it consistent with leveldb's iterator and maybe use `strings.HasPrefix` as proposed in #181. 
And fix the comment of course. Would you agree on that?

mrahs:
Agreed. When I surveyed the implementations (memory, redis, merged, leveldb), it looked like all of them were compatible with calling `Next` after a `Seek` except for leveldb's.

I agree with your reasoning. Let's fix the comment and the memory iterator.
I can rework my PR sometime within the next couple of weeks.

frairon:
Wait what? But if all are compatible with calling `next` after `seek`, that means that we need to allow calling `Next` in leveldb too, so should rather use the current PR?

mrahs:
It comes down to memory vs leveldb. Redis' seek is a noop. The merged iterator is primarily used with leveldb. And I'm getting the impression that LevelDB is probably the only one used in production.

Given that, I'd rather avoid a hack in production code. Thoughts?

frairon:
Yep, agreeing to that. So let's not use a hack, fix the comment and memory iterator. Just got a bit confused by your earlier comment, because it sounded that "next" after "seek" works, but it doesn't without losing one element.

frairon:
@mrahs, while you're on it, could you maybe also replace the type ` header.Headers` in the memory-storage (and in its iterator) [here](https://github.com/lovoo/goka/blob/b037ca1e4d7a58ba95ea70b8c143091891828f0d/storage/memory.go#L71) with a raw data type like `map[string][]byte`? 
I think the headers type added their by mistake by find/replace when introducing the header-type, but it logically does not belong there.

mrahs:
My bad 🤦
Will definitely do.

mrahs:
Updated the PR to resolve multiple issues since they are all related: #318, #331, and #181.

options.go
```diff
@@ -15,7 +15,7 @@ import (
 // UpdateCallback is invoked upon arrival of a message for a table partition.
 // The partition storage shall be updated in the callback.
 // Headers can converted to a map using FromSarama function.
-type UpdateCallback func(s storage.Storage, partition int32, key string, value []byte, headers ...*sarama.RecordHeader) error
+type UpdateCallback func(s storage.Storage, partition int32, key string, value []byte, headers Headers) error
```
frairon, line 18: Thanks, awesome! Only thing is, that we're now converting the headers on all message, even though we might never use them. If we'd have a new context that wraps the `sarama.ConsumerMessage` passed to the update callback
```go

type UpdateContext interface{
   Headers() Headers
}

// then pass this instead of the headers
type UpdateCallback func(s storage.Storage, partition int32, key string, value []byte, updCtx UpdateContext) error
```
then we could transform the headers only if we need them. Also we would be flexible in case we want to add more fields at some point, like Offset or Timestamp and not break the whole API by just adding a new function in the interface.

But of course it's more work and I don't want to load that on you and I'm happy to support you there :). 
Still though - do you think it's worth it or is the use case so hypothetical that we can live with the performance-overhead (if any) and the extra API-change and leave it as is?

mrahs: Good catch!

I like the idea of introducing `UpdateContext` considering this is already a breaking change. I think it makes sense to wrap other arguments in the context as well:
```go
type UpdateContext interface {
  Storage() storage.Storage
  Key() string
  Value() []byte
  Partition() int32
  Headers() Headers
}

type UpdateCallback func(ctx UpdateContext)
```

Thoughts?

frairon: Awesome, love it. Could even also "Topic()", since the rest of the fields are redundant too.
Many thanks for your work!

mrahs: Thanks!
Done.

options.go
```diff
@@ -9,13 +9,40 @@ import (
...
+	// when headers are not used.
+	Headers() Headers
+}
+
 // UpdateCallback is invoked upon arrival of a message for a table partition.
-// The partition storage shall be updated in the callback.
-// Headers can converted to a map using FromSarama function.
-type UpdateCallback func(s storage.Storage, partition int32, key string, value []byte, headers ...*sarama.RecordHeader) error
+// The partition storage in the context shall be updated in the callback.
+type UpdateCallback func(ctx UpdateContext) error
```
frairon, line 45: I hate to be the picky/pedantic douche here, but to be real honest, moving all arguments into the single interface looks a bit strange to me. 
The use case of the update callback is in 99% of the cases: execute default i.e. store `key`/`value` in `storage`. 
So I think those three values (+ the partition for backwards compatibility) should be part of the callback function. The rest is sugar on top and only rarely used, if someone needs the headers and whatnot, so it can be wrapped in the object. 
Having all default values hidden behind functions also adds three extra function calls, that cannot be inlined. The performance overhead is probably not measurable as the storage will be much slower, but still it seems like an unnecessary penalty.
In my head, the type would look like this:
`type UpdateCallback func(s storage.Storage, partition int32, key string, value []byte, ctx UpdateContext) error`
(maybe the linter will complain that `ctx` is not the first argument, then we could rename it).
What do you think? 
Still, many thanks for your work here and sorry for the pickyness. Let me know if you need support.

partition_table.go
```diff
@@ -521,7 +523,7 @@ func (p *PartitionTable) loadMessages(ctx context.Context, cons sarama.Partition
 			}
 
 			lastMessage = time.Now()
-			if err := p.storeEvent(string(msg.Key), msg.Value, msg.Offset, msg.Headers...); err != nil {
+			if err := p.storeEvent(string(msg.Key), msg.Value, msg.Offset, headers.FromSarama(msg.Headers)); err != nil {
```
frairon, line 526: this will be called on every message being loaded, but actually we wanted to avoid converting all headers, right? This makes the lazyloading implemented in the UpdateContext kind of useless, as the field `saramaHeaders` is never actually instantiated, or am I missing something?

mrahs: _sigh_

Thank you for stopping me from merging suboptimal code. I should have been more thorough with this PR. I'll fix those asap.

options.go
mrahs: Yep, the performance overhead is negligible in the context of IO. This is similar to the processor callback context which is probably executed much more often in the wild, and might not be using persistence which means it's more likely for any overhead to manifest in that path.

I see your point about optimizing for the main usecase, but having both callbacks follow the same pattern is a good argument as well.

Thoughts?

frairon: Not convinced yet :) 
The performance was actually not my point, shouldn't have brought it up. The main point was that the function's purpose is to store key/value in storage, so those three things should be part of the arguments and not hidden behind some context. 
This would also match the processor-callback scheme, which contains the centric item (the message) directly in the parameters, not in the context: `func(ctx goka.Context, msg interface{})`.

Then the default is simple:
```go
func update(s storage.Strorage, key string, value []byte, ctx updateContext){
    s.Set(key, value)
}
```
Which makes semantically much more sense to me than
```go
func update(ctx updateContext){
    ctx.Storage().Set(ctx.Key(), ctx.Value())
}
```
Don't you think?

I mean we wouldn't write a sum function like this
```go
type ops interface{
 A() float64
 B() float64
 SetResult(float64)
}
func sum(values ops){
    ops.SetResult(ops.A()+ops.B())
}
```
Right?
We'd simply write
```go
func sum(a, b float64)float64{
    return a+b
}
```

I'd also leave the partitionID argument, so basically leave the function as it was before, just replace the `sarama.Headers` with the generic `UpdateContext`, to minimize the breaking changes.
What do you think?

mrahs: Good point Franz. I'm convinced by the centric-item argument :)

Considering we are making a breaking change, Reordering the arguments to `UpdateContext, Storage, Key, Value` and moving partition to `UpdateContext` (since it sounds like it's rarely used) looks more idiomatic go to me. The callback might get `conext.Context` at some point which should be expected to be (within) the first argument.

```go
func update(ctx UpdateContext, s storage.Strorage, key string, value []byte){
    s.Set(key, value)
}
```
Do I have your vote on that one?

frairon: The reordering and removing partition makes total sense.
Love it :heart:

frairon:
@mrahs, would you by any chance have any progress on the PR?
We're kind of waiting for that so we can merge other features and release the breaking changes together. But no pressure :).
No seriously, I don't want to stress you in any way. As an alternative, we could also merge this PR into a feature-branch in goka and I'll implement the remaining issues we aggreed upon, then we merge everything into master when it's done. I just don't want to steal that from you.

mrahs:
> @mrahs, would you by any chance have any progress on the PR?
> We're kind of waiting for that so we can merge other features and release the breaking changes together. But no pressure :).
> No seriously, I don't want to stress you in any way. As an alternative, we could also merge this PR into a feature-branch in goka and I'll implement the remaining issues we aggreed upon, then we merge everything into master when it's done. I just don't want to steal that from you. What do you think?

Sorry for the silence. I've been meaning to wrap this up asap, but life got busy. I'll update the PR later today.

options.go
mrahs: Done.

frairon (dismissed):
Looks good, many thanks for the extra changes. Your contribution is really appreciated here!!
I think we can leave it like this, so we can go forward :).

One question though: the conversion between sarama headers and goka-headers is now moved to its own structure. But the only two places where it's being stored for lazy-conversion is the `cbContext` and the `UpdateContext`, whereas both could also just store the sarama.Headers and convert on request.
It just seems a bit of overkill to me to have this rather complex `Headers` structure supporting both header-types, with things like `eval`, `Len`, `ToSaramaPtr`, `ToSarama` and all that, which is mostly needed for testing, instead of the simpler way like before. So I'm just wondering if I'm missing the bigger picture here why this complexity is needed? :) 

But again, thanks a lot for your work and your time on all that!

mrahs:
> Looks good, many thanks for the extra changes. Your contribution is really appreciated here!!
> I think we can leave it like this, so we can go forward :).
> 
> One question though: the conversion between sarama headers and goka-headers is now moved to its own structure. But the only two places where it's being stored for lazy-conversion is the `cbContext` and the `UpdateContext`, whereas both could also just store the sarama.Headers and convert on request.
> It just seems a bit of overkill to me to have this rather complex `Headers` structure supporting both header-types, with things like `eval`, `Len`, `ToSaramaPtr`, `ToSarama` and all that, which is mostly needed for testing, instead of the simpler way like before. So I'm just wondering if I'm missing the bigger picture here why this complexity is needed? :)
> 
> But again, thanks a lot for your work and your time on all that!

Thanks Franz!
It sounds like you're not fully content with the code. I'm open to further discussions.

The last iteration attempts to satisfy the following requirements and constraints:
- Avoid forcing users to import Sarama as a dependency. Using a custom header type rather than exposing Sarama's achieves this. It was first introduced in [this commit](https://github.com/lovoo/goka/commit/373b87deda5fc209d80cddfe9a3dc559d0d0b1e4#diff-4eec45b6a922746411c0f8aea183e30f77af1b0a5bea1688eaf343a5f9c23703) with a type that matches what [Apache uses for its implementation](https://github.com/apache/kafka/blob/22541361b7c62913bc848fb03e5f807e152a1023/clients/src/main/java/org/apache/kafka/common/header/Headers.java#L38).
- [Sarama's `ProducerMessage` type](https://github.com/lovoo/goka/blob/e51ca796a29cdd8e5fe9ee130d6c8a42ff7efc2b/producer.go#L86) takes a slice of values while [its `ConsumerMessage` type](https://github.com/lovoo/goka/blob/e51ca796a29cdd8e5fe9ee130d6c8a42ff7efc2b/tester/consumer.go#L111) takes a slice of pointers. Conversion functions provide the convenience of switching between these formats. 
- Optimize for processors that don't make use of headers, but make headers readily available for processors that need them. Lazy evaluation achieves this.

Now all of this can still be achieved with a simpler header type of `type Headers map[string][]byte` if it weren't for the storage proxy which has a [public function with headers in the argument list](https://github.com/lovoo/goka/blob/e51ca796a29cdd8e5fe9ee130d6c8a42ff7efc2b/proxy.go#L32). Keeping it as the sarama type forces implementers to import sarama. Switching to the goka type requires more complexity for lazy evaluation.

Now that I've typed this, I believe it should be possible to change the proxy interface to use the `UpdateContext` type which would achieve both goals (no external dependency and lazy evaluation). I'm thinking:
```go
func (s *storageProxy) Update(ctx UpdateContext, k string, v []byte) error {
  return s.update(ctx, s, k, v)
}
```
I've just tired this out and it [looks better](https://github.com/lovoo/goka/compare/master...mrahs:headers-refactor?expand=1). The headers are `type Headers map[string][]byte` with only `Merged`, `ToSarama` and `ToSaramaPtr` functions. I can push this commit if that looks good to you.

frairon:
Awesome, that looks even better! 
Funnily after writing my comment I also thought your solution is fine and maybe the encapsulation in the extra structure isn't such a bad idea. Also since you're the main contributor around headers you probably know better what seems more elegant there. But you were faster responding.
Anyway, if you ask me this new solution is even better and we can push that. Thanks also for removing the unused fields and functions in `DefaultUpdateContext` :).

frairon (approved):

PR #324, opened 2021-05-17
Expose headers via the root package

### LINKED ISSUE #322: Expose headers interface via the root package
Just like `goka.Codec`, it would be nice to have `goka.Headers` rather than `headers.Headers`.

mrahs:
No functional or breaking changes. Mostly a find/replace exercise. The meat of it is the ability for users to write `goka.Headers` instead of `headers.Headers` if they choose to.

The choice of implementing it using a type alias is two folds:
- There is little value in introducing an `interface` type definition to the mix since users are not expected to provide implementations.
- Introducing the interface type alongside the struct type `headers.Headers` requires some quirks to avoid cyclic imports between `goka` and `headers` packages. Not worth it.

Closes #322

### DIFF
--- a/systemtest/processor_test.go
+++ b/systemtest/processor_test.go
@@ -289,6 +289,10 @@ func TestRecoverAhead(t *testing.T) {
 // TestRebalance runs some processors to test rebalance. It's merely a
 // runs-without-errors test, not a real functional test.
 func TestRebalance(t *testing.T) {
+	if !*systemtest {
+		t.Skipf("Ignoring systemtest. pass '-args -systemtest' to `go test` to include them")
+	}
+
 	var (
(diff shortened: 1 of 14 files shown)


systemtest/processor_test.go
```diff
@@ -289,6 +289,10 @@ func TestRecoverAhead(t *testing.T) {
 // TestRebalance runs some processors to test rebalance. It's merely a
 // runs-without-errors test, not a real functional test.
 func TestRebalance(t *testing.T) {
+	if !*systemtest {
```
frairon, line 292: damn :facepalm:  thanks for that!

frairon (approved):

juandes (approved):

PR #327, opened 2021-05-28
Added GetTableKeys to tester

juandes:
This PR adds a function `GetTableKeys` that returns a tester Table's list of keys.

### DIFF
--- a/tester/tester.go
+++ b/tester/tester.go
@@ -383,6 +383,26 @@ func (tt *Tester) waitForClients() {
 	logger.Printf("waiting for consumers done")
 }
 
+// GetTableKeys returns a Table's keys.
+func (tt *Tester) GetTableKeys(table goka.Table) []string {
+	tt.mStorages.Lock()
+	defer tt.mStorages.Unlock()
+
+	keys := make([]string, 0)
+	topic := string(table)
+	st, exists := tt.storages[topic]
+	if !exists {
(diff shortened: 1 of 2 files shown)


tester/tester.go
```diff
@@ -383,6 +383,26 @@ func (tt *Tester) waitForClients() {
...
+// GetTableKeys returns a Table's keys.
+func (tt *Tester) GetTableKeys(table goka.Table) []string {
+	tt.mStorages.Lock()
+	defer tt.mStorages.Unlock()
+
+	keys := make([]string, 0)
+	topic := string(table)
+	st, exists := tt.storages[topic]
+	if !exists {
+		panic(fmt.Errorf("topic %s does not exist", topic))
```
frairon, line 395: you don't have to panic, you can actually use the `testing.T` that is added when creating the Tester: `tt.t.Fatalf(...)`

tester/tester.go
```diff
@@ -383,6 +383,26 @@ func (tt *Tester) waitForClients() {
...
+	}
+
+	it, _ := st.Iterator()
```
frairon, line 398: `defer it.Release()`  is missing. I know it's just a tester, but who knows how our implementation might change in future and we require the iterators to be released...?

juandes: Fair point.

PR #338, opened 2021-07-14
NewTopicManagerConfig doesn't seem to match its own docstring

mikedewar:
changed NewTopicManagerConfig to match its docstring, with replication factors both set to 1


mikedewar:
oop this is a duplicate; closing.

PR #339, opened 2021-07-22
offset fix

frairon:


processor.go
```diff
@@ -612,7 +612,7 @@ func (g *Processor) Setup(session sarama.ConsumerGroupSession) error {
 		// create partition processor for our partition
 		pproc, err := g.createPartitionProcessor(session.Context(), partition, runModeActive,
 			func(msg *message, metadata string) {
-				session.MarkOffset(msg.topic, msg.partition, msg.offset, metadata)
+				session.MarkOffset(msg.topic, msg.partition, msg.offset+1, metadata)
```
ubunatic, line 615: Please leave a comment why a +1 is required here. It look very arbitrary to the average user.

systemtest/processor_test.go
```diff
@@ -430,3 +431,79 @@ func TestCallbackFail(t *testing.T) {
...
+			err := proc.Run(ctx)
+			test.AssertNil(t, err)
+		}()
+
+		time.Sleep(10 * time.Second)
+		val, err := proc.Get("1")
+		test.AssertNil(t, err)
+		test.AssertTrue(t, val != nil)
+		test.AssertEqual(t, val.(int64), int64(10))
+		log.Printf("received 10 messages")
```
jomaresch, line 504: Sorry, but i don't know how this tests the commit of the input topic. 
It looks like it only tests the `ctx.SetValue`, but not the commit of the input topic. 
Maybe i miss something?

processor.go
ubunatic: FYI: [tester](https://github.com/lovoo/goka/blob/0085fa7f7c2e542e740ec565af7662b2d441b019/tester/consumergroup.go#L262) vs. [sarama](https://github.com/Shopify/sarama/blob/master/consumer_group.go#L642)
```
// tester
func (cgs *cgSession) MarkMessage(msg *sarama.ConsumerMessage, metadata string) {
	cgs.MarkOffset(msg.Topic, msg.Partition, msg.Offset, metadata)
}
```
```
// sarama
func (s *consumerGroupSession) MarkMessage(msg *ConsumerMessage, metadata string) {
	s.MarkOffset(msg.Topic, msg.Partition, msg.Offset+1, metadata)
}
```
:thinking:

ubunatic: Maybe we should use MarkMessage instead of MarkOffset?

frairon: oh good point regarding the tester. However we cannot use MarkMessage, because the whole point of the original change was to get rid of the sarama-message and have an own message type. But MarkMessage requires the sarama-message

systemtest/processor_test.go
frairon: sorry this lacks some description, was a bit in a hurry :) So basically the processor just receives and accumulates the value. There's only one key. Now the idea was that we restart the same processor a couple of times and check the value of the one key we ever store. It should never change, because the first processor run processes all messages and there aren't any more.
Without the "+1"-fix, it keeps increasing with each restart of the processor (and the test fails), as the last message is being reprocessed every time, thus incrementing the persisted value.
I'll add something like this as a comment.

processor.go
frairon: and a comment is definitely necessary, will do!

ubunatic: In the comment, you may just refer to the sarama code that we are adapting here. Maybe they have a similar issue/discussion/PR where this is already explained?

ubunatic (approved):
The code is correct, even though we agree to have some more comments.
I would approve already it to not block this bug fix for our other projects.

systemtest/processor_test.go
jomaresch: Right. Thanks for the explanation 👍🏼

PR #341, opened 2021-08-10
Visitor Tool implementation

frairon:


examples/10-visit/main.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+func main() {
+
+	tm, err := goka.NewTopicManager(brokers, goka.DefaultConfig(), tmc)
+	if err != nil {
+		log.Fatalf("Error creating topic manager: %v", err)
+	}
+	err = tm.EnsureStreamExists(string(topic), 8)
```
mmreza79, line 62: why it is 8? we have 20 partitions, right?

examples/8-monitoring/main.go
```diff
@@ -99,7 +101,7 @@ func process(ctx goka.Context, msg interface{}) {
 
 	u.Clicks++
 	ctx.SetValue(u)
-	fmt.Printf("[proc] key: %s clicks: %d, msg: %v\n", ctx.Key(), u.Clicks, msg)
+	// fmt.Printf("[proc] key: %s clicks: %d, msg: %v\n", ctx.Key(), u.Clicks, msg)
```
mmreza79, line 104: don't you want to remove this line?

integrationtest/processor_test.go
```diff
@@ -169,6 +169,60 @@ func TestHeaders(t *testing.T) {
...
+	test.AssertEqual(t, k, "a")
+	test.AssertEqual(t, v, int64(15))
+	test.AssertEqual(t, gkt.TableValue("test-table", "a"), int64(15))
+
+	cancel()
+	<-done
+	test.AssertNil(t, err)
+}
+
 /*
```
mmreza79, line 226: following commented code should remove. right?

examples/10-visit/main.go
frairon: that's a completely arbitrary number here, can be 8 or 20 or 100. It's just an example how to use the visitor-tool. This should work with any number of partitions

integrationtest/processor_test.go
frairon: well I mean the commented code was there before this PR, I thought I'd leave to get a reminder that there are still some things missing to be tested.. But we can also remove it, that's true.

partition_processor.go
```diff
@@ -589,3 +645,43 @@ func (pp *PartitionProcessor) processMessage(ctx context.Context, wg *sync.WaitG
 	msgContext.finish(nil)
 	return nil
 }
+
+// VisitValues iterates over all values in the table and calls the "visit"-callback for the passed name.
+// Optional parameter value can be set, which will just be forwarded to the visitor-function
+func (pp *PartitionProcessor) VisitValues(ctx context.Context, name string, wg *sync.WaitGroup, meta interface{}) error {
+	if pp.table == nil {
+		return fmt.Errorf("cannot visiit values in stateless processor")
```
mmreza79, line 653: (typo) cannot visit...

examples/8-monitoring/main.go
frairon: nah, I'd rather actually comment it in again so we get the print. It's an example after all, so it does not really matter if the print is here or not.

web/actions/server.go
```diff
@@ -0,0 +1,223 @@
...
+		"page_title": "Actions",
+		"actions":    s.sortedActions(),
+		"error":      r.URL.Query()["error"],
+	}
+
+	s.executeTemplate(w, params)
+}
+
+// func (s *Server) source(w http.ResponseWriter, r *http.Request) {
+// 	vars := mux.Vars(r)
```
mmreza79, line 155: This commented code should remove. right?

integrationtest/processor_test.go
frairon: actually I'd like to keep it, so we can see what tests we used to have and maybe migrate some.

examples/8-monitoring/main.go
frairon: commented it in again.

partition_processor.go
frairon: fixed, thanks!

mmreza79 (approved):
Too big PR :) but look good to me.

processor.go
```diff
@@ -609,7 +609,13 @@ func (g *Processor) Setup(session sarama.ConsumerGroupSession) error {
 	// create partition views for all partitions
 	for partition := range assignment {
 		// create partition processor for our partition
-		pproc, err := g.createPartitionProcessor(session.Context(), partition, runModeActive, session.MarkMessage)
+		pproc, err := g.createPartitionProcessor(session.Context(), partition, runModeActive,
+			func(msg *message, metadata string) {
+				// We have to commit the offset that we want to read next, not the one that we just have
+				// processed, therefore msg.offset+1 is required to get the next message.
+				// This has the same semantics as sarama's implementation of session.MarkMessage (which calls MarkOffset with offset+1)
+				session.MarkOffset(msg.topic, msg.partition, msg.offset+1, metadata)
```
ubunatic, line 617: I still do not like the `+1` here in the processor. This is an offset-management detail and should be wrapped if possible.

tester/consumergroup.go
```diff
@@ -237,15 +237,15 @@ func (cgs *cgSession) MarkOffset(topic string, partition int32, offset int64, me
 	cgs.mMessages.Lock()
 	defer cgs.mMessages.Unlock()
 
-	msgKey := cgs.makeMsgKey(topic, offset)
+	msgKey := cgs.makeMsgKey(topic, offset-1)
```
ubunatic, line 240: Maybe add a panic on `offset < 1`

web/actions/action.go
```diff
@@ -0,0 +1,85 @@
...
+func (a *action) Start(value string) {
+	a.Lock()
+	defer a.Unlock()
+	a.started = time.Now()
+	a.finished = time.Time{}
+	ctx, cancel := context.WithCancel(context.Background())
+	a.ctx, a.cancel = ctx, cancel
+	go func() {
+		defer cancel()
+		err := a.actor.RunAction(ctx, value)
```
ubunatic, line 71: this `a` access is outside of the locks, I guess `actor` does not need to be protected?

mockautoconsumers.go
```diff
@@ -505,17 +505,17 @@ func (cg *MockConsumerGroup) topicKey(topics []string) string {
 	return strings.Join(topics, ",")
 }
 
-func (cg *MockConsumerGroup) markMessage(msg *sarama.ConsumerMessage) {
+func (cg *MockConsumerGroup) markMessage(topic string, partition int32, offset int64, metadata string) {
 	cg.mMessages.Lock()
 	defer cg.mMessages.Unlock()
 
-	cnt := cg.messages[msg.Offset]
+	cnt := cg.messages[offset-1]
```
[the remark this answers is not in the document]
frairon: @ubunatic it's not the offset value that is being checked here, but whether a message with that offset exists. If it doesn't exist (cnt==0), then it panics, because the consumer is trying to "commit" a message with an offset, that it never received.

mockautoconsumers.go
```diff
@@ -505,17 +505,17 @@ func (cg *MockConsumerGroup) topicKey(topics []string) string {
...
 
 	if cnt == 0 {
-		panic(fmt.Errorf("Cannot mark message with offest %d, it's not a valid offset or was already marked", msg.Offset))
+		panic(fmt.Errorf("Cannot mark message with offest %d, it's not a valid offset or was already marked", offset))
```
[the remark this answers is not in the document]
frairon: fixed

partition_processor.go
```diff
@@ -386,15 +398,21 @@ func (pp *PartitionProcessor) run(ctx context.Context) (rerr error) {
...
 			return
-
+		case visit, open := <-pp.visitInput:
+			if !open {
+				return nil
+			}
+			err := pp.processVisit(ctx, &wg, visit, syncFailer, asyncFailer)
+			if err != nil {
+				return fmt.Errorf("Error migrating %s for %s: %v", visit.name, visit.key, err)
+			}
```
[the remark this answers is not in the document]
frairon: thanks, fixed

partition_processor.go
```diff
@@ -528,7 +546,45 @@ func (pp *PartitionProcessor) enqueueTrackOutputStats(ctx context.Context, topic
...
 }
 
-func (pp *PartitionProcessor) processMessage(ctx context.Context, wg *sync.WaitGroup, msg *sarama.ConsumerMessage, syncFailer func(err error), asyncFailer func(err error)) error {
+func (pp *PartitionProcessor) processVisit(ctx context.Context, wg *sync.WaitGroup, v *visit, syncFailer func(err error), asyncFailer func(err error)) error {
+	cb, ok := pp.visitCallbacks[v.name]
+	// no callback registered for migration
+	if !ok {
+		return fmt.Errorf("no callback registered for migration")
+	}
+
```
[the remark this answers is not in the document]
frairon: fixed, this was just legacy since the feature was earlier called "migration" based on our main use case. Just later we decided to rename it to "visit" as it's more generic.

partition_processor.go
```diff
@@ -589,3 +645,43 @@ func (pp *PartitionProcessor) processMessage(ctx context.Context, wg *sync.WaitG
...
+			// context is closed, so we know that the current visit never entered the channel
+			// so we have to remove it from the waitgroup to avoid the lock.
+			wg.Done()
+			return nil
+		// enqueue the visit
+		case pp.visitInput <- &visit{
+			key:  string(it.Key()),
+			name: name,
+			meta: meta,
+			done: wg.Done,
```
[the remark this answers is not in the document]
frairon: @ubunatic
the thing is: it cannot be called concurrently, that's the point of the `select`. If we get a `ctx.Done()`, then the message won't be put into `pp.visitInput`, so it won't be processed by the visitor, so it will never call `wg.Done()` there. In the other case that it ends up in the channel, then it won't go  for the `ctx.Done()` case anymore. 
So actually it should be fine.

processor.go
frairon: I tried to make it more clear in the next PR

tester/consumergroup.go
frairon: but this is totally tester internal specific? the offset is generated by the tester and it's also only used by the tester.

web/actions/action.go
```diff
@@ -0,0 +1,85 @@
...
+// so starting it a second time might lead to overwriting the finish time if the first run
+// stops and writes "finish" again
+func (a *action) Start(value string) {
+	a.Lock()
+	defer a.Unlock()
+	a.started = time.Now()
+	a.finished = time.Time{}
+	ctx, cancel := context.WithCancel(context.Background())
+	a.ctx, a.cancel = ctx, cancel
+	go func() {
```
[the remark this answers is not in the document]
frairon: true that, but relying on the server is actually also bad design, I'll try to improve it in the next PR

web/actions/action.go
frairon: yep, the lock only protects variables that can change during lifetime of `action`, and `actor` isn't one of them, so doesn't need protection.

examples/10-visit/main.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		log.Printf("Error creating kafka topic %s: %v", topic, err)
+	}
```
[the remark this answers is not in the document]
frairon: done.

PR #345, opened 2021-09-13
fixed comments from PR #341, added unit tests for action

frairon:


processor.go
```diff
@@ -974,3 +970,18 @@ func ensureCopartitioned(tm TopicManager, topics []string) (int, error) {
 	}
 	return npar, nil
 }
+
+// createMessageCommitter returns a commitCallback that allows to commit a message into the passed
+// sarama.ConsumerGroupSession.
+//
+// Note that the offset to be committed must be offset that the consumer expects to consume next, not the offset of the message.
```
ubunatic, line 977:
```suggestion
// Note that the offset to be committed must be the offset that the consumer expects to consume next, not the offset of the message.
```

processor.go
```diff
@@ -974,3 +970,18 @@ func ensureCopartitioned(tm TopicManager, topics []string) (int, error) {
...
+// See documentation at https://kafka.apache.org/25/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html which says:
+//
+//   Note: The committed offset should always be the offset of the next message that your application will read. Thus, when calling commitSync(offsets) you should add one to the offset of the last message processed.
+//
+// This has the same semantics as sarama's implementation of session.MarkMessage (which calls MarkOffset with offset+1)
```
ubunatic, line 982: Thanks for the details, I think they are critical to document here. :+1:

ubunatic (approved):
LGTM

PR #348, opened 2021-09-20
topic manager wait

frairon:
This PR adds a simple polling mechanism for the TopicManager after creating a topic to be fully populated in the cluster.

### DIFF
--- a/topic_manager.go
+++ b/topic_manager.go
@@ -230,9 +230,14 @@ func (m *topicManager) ensureExists(topic string, npar, rfactor int, config map[
 }
 
 func (m *topicManager) waitForCreated(topic string) error {
-	start := time.Now()
-	for m.topicManagerConfig.CreateTopicTimeout != 0 {
 
+	// no timeout defined -> no check
+	if m.topicManagerConfig.CreateTopicTimeout == 0 {
+		return nil
+	}
+
+	start := time.Now()
+	for {
 		// past timeout, break the loop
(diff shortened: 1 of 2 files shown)


topic_manager.go
```diff
@@ -229,6 +229,38 @@ func (m *topicManager) ensureExists(topic string, npar, rfactor int, config map[
...
 }
 
+func (m *topicManager) waitForCreated(topic string) error {
+
+	// no timeout defined -> no check
+	if m.topicManagerConfig.CreateTopicTimeout == 0 {
+		return nil
+	}
+
+	start := time.Now()
```
paulhenke, line 239: Another approach
```
// or better get ctx passed as param in waitForCreated instead of using .Background()
ctx, cancel := context.WithTimeout(context.Background, m.topicManagerConfig.CreateTopicTimeout)
defer cancel()

for ctx.Err() == nil {
	_, err := m.Partitions(topic)
	switch err {
	case nil: 
		return nil
	case errTopicNotFound:
		time.Sleep(time.Second)
	default: 
		return fmt.Errorf("error checking topic: %w", err)
	}
}
```

frairon: I was thinking of adding the context as a param, but that would mean changing the whole api.
But the timeout context is definitely an improvement, didn't think of this. so Thanks :)

frairon: @paulhenke fixed.

jomaresch (approved):

PR #353, opened 2021-09-29
improving processor error handling/shutdown

frairon:
This PR fixes two main things:
* Refactoring of error handling to shutdown gracefully or reconnect 
* replace multierr.Errors with hashicorp/go-multierror


partition_processor.go
```diff
@@ -297,50 +294,58 @@ func (pp *PartitionProcessor) Start(ctx context.Context) error {
...
+}
+
 // Stop stops the partition processor
 func (pp *PartitionProcessor) Stop() error {
 	pp.log.Debugf("Stopping")
 	defer pp.log.Debugf("... Stopping done")
 	pp.state.SetState(PPStateStopping)
-	defer pp.state.SetState(PPStateIdle)
-	errs := new(multierr.Errors)
+	defer pp.state.SetState(PPStateStopped)
```
ubunatic, line 306: Idle will never be set again. Is this intended?

partition_processor.go
```diff
@@ -297,50 +294,58 @@ func (pp *PartitionProcessor) Start(ctx context.Context) error {
...
 
 func (pp *PartitionProcessor) run(ctx context.Context) (rerr error) {
 	pp.log.Debugf("starting")
 	defer pp.log.Debugf("stopped")
 
-	errs := new(multierr.Errors)
+	// protect the errors-collection
+	var mutexErr sync.Mutex
+
 	defer func() {
```
ubunatic, line 345: The mutex is used a lot in this long func. You could add a `setOrAddError` func that has uses the mutex exclusively and use this func in all places where you want to set or add errors.

This func could then also do a nil check and append if not nil to ensure no errors are lost.

processor.go
```diff
@@ -352,56 +336,50 @@ func (g *Processor) rebalanceLoop(ctx context.Context, consumerGroup sarama.Cons
...
 
 	defer func() {
-		errs.Collect(rerr)
-		rerr = errs.NilOrError()
+		rerr = multierror.Append(errs, rerr).ErrorOrNil()
 	}()
 
 	defer func() {
 		g.log.Debugf("closing consumer group")
 		defer g.log.Debugf("closing consumer group ... done")
```
ubunatic, line 347: the nested defer and the many defers look like a code-smell (not sure though).
Go will unroll all defers from bottom to top anyway. Do you need the nesting at all?
Maybe you can also group all teardown logs in one defer func?

partition_processor.go
```diff
@@ -71,7 +74,6 @@ type PartitionProcessor struct {
 	// runnerErrors store the errors occuring during runtime of the
 	// partition processor. It is created in Setup and after the runnerGroup
 	// finishes.
-	runnerErrors chan error
```
jomaresch, line 74: we should also remove the comment

partition_processor.go
```diff
@@ -188,33 +190,24 @@ func (pp *PartitionProcessor) Recovered() bool {
...
-	return pp.runnerErrors
-}
-
 // Start initializes the partition processor
 // * recover the table
 // * recover all join tables
 // * run the join-tables in catchup mode
 // * start the processor processing loop to receive messages
-func (pp *PartitionProcessor) Start(ctx context.Context) error {
+func (pp *PartitionProcessor) Start(setupCtx, ctx context.Context) error {
```
jomaresch, line 198: maybe a comment would be nice to explain the difference between the `setupCtx` and `ctx`, because it's quiet unusual to have two ctxs.

errors.go
```diff
@@ -35,19 +35,34 @@ var (
...
+// the message, e.g. panic, (de|en)-coding errors or invalid usage of context.
+type errSetup struct {
+	partition int32
+	err       error
+}
+
+func (ec *errSetup) Error() string {
+	return fmt.Sprintf("error setting up (partition=%d): %v", ec.partition, ec.err)
+}
+func newErrSetup(partition int32, err error) error {
```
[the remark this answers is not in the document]
frairon: hmmm, for this case it didn't, cleaned up multiple newlines though. fixed it.

errors.go
```diff
@@ -35,19 +35,34 @@ var (
...
+		partition: partition,
+		err:       err,
+	}
 }
 
-func newErrProcessing(err error) error {
-	return &errProcessing{
-		wrapped: err,
+// type to indicate that some non-transient error occurred while processing
+// the message, e.g. panic, (de|en)-coding errors or invalid usage of context.
```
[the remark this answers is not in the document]
frairon: actually the comment is wrong in this place, because this describes a setup-error (which cannot be a codec error).
But I fixed it in the other place and rewrite this comment.

partition_processor.go
```diff
@@ -188,33 +190,24 @@ func (pp *PartitionProcessor) Recovered() bool {
...
+
+	if state := pp.state.State(); state != PPStateIdle {
+		return fmt.Errorf("partitionprocessor seems not idle (but %v), cannot start", state)
```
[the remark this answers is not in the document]
frairon: true :D

partition_processor.go
frairon: it's set at the beginning as an initial state. But once it leaves the idle state it will never go back to avoid reusing a partition-processor. It's a run-once-thing, that's just safer to implement that way.

partition_processor.go
```diff
@@ -350,7 +355,9 @@ func (pp *PartitionProcessor) run(ctx context.Context) (rerr error) {
 			// only fail processor if context not already Done
 			select {
 			case <-ctx.Done():
-				rerr = err
+				mutexErr.Lock()
+				rerr = newErrProcessing(pp.partition, fmt.Errorf("synchronous error in callback: %v", err))
```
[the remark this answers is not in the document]
frairon: Nice catch. Technically it's quite likely that syncfailers will always be executed before, but it could happen that the asyncfailer fails first (e.g. if the consume-callback has a sleep?)
Anyway, it's fixed.

partition_processor.go
```diff
@@ -372,7 +381,9 @@ func (pp *PartitionProcessor) run(ctx context.Context) (rerr error) {
 
 	defer func() {
 		if r := recover(); r != nil {
-			rerr = fmt.Errorf("%v\n%v", r, strings.Join(userStacktrace(), "\n"))
+			mutexErr.Lock()
+			rerr = newErrProcessing(pp.partition, fmt.Errorf("panic in callback: %v\n%v", r, strings.Join(userStacktrace(), "\n")))
```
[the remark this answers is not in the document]
frairon: same here, very very unlikely but I can't rule it out, so let's append it.

processor.go
frairon: yeah the nesting looks a bit ugly here, but the outer defer is necessary since the loop below never breaks (but returns). But I removed the inner defer to make it look less ugly.

partition_processor.go
```diff
@@ -363,7 +370,9 @@ func (pp *PartitionProcessor) run(ctx context.Context) (rerr error) {
 		// asyncFailer is called asynchronously from other goroutines, e.g.
 		// when the promise of a Emit (using a producer internally) fails
```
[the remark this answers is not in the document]
frairon: fixed

errors.go
```diff
@@ -35,19 +35,34 @@ var (
...
+type errSetup struct {
+	partition int32
+	err       error
+}
```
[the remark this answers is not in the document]
frairon: jep, just fixed it :)

mockautoconsumers.go
```diff
@@ -551,18 +552,16 @@ func (cg *MockConsumerGroup) Consume(ctx context.Context, topics []string, handl
...
-		errs.Collect(errg.Wait().NilOrError())
-
-		// cleanup and collect errors
-		errs.Collect(handler.Cleanup(session))
+		err = multierror.Append(
+			// wait for runner errors and collect error
+			errg.Wait().ErrorOrNil(),
+			// cleanup and collect errors
+			handler.Cleanup(session),
+		)
```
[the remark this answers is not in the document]
frairon: good catch! fixed

partition_processor.go
frairon: damn, that happens if you edit your code in an external diff-editor, not the actual ide :facepalm:

partition_processor.go
frairon: totally agree, and it still messes with my head. Added a comment.

processor.go
```diff
@@ -314,7 +284,7 @@ func (g *Processor) Run(ctx context.Context) (rerr error) {
 		g.log.Debugf("closing producer")
 		defer g.log.Debugf("producer ... closed")
 		if err := g.producer.Close(); err != nil {
-			merrors.Collect(fmt.Errorf("error closing producer: %v", err))
+			errs = multierror.Append(errs, fmt.Errorf("error closing producer: %w", err))
```
ubunatic, line 287: Setting the `errs` here in this deferred func and then later using the `errs` to fill the `rerr` in the previous (but later called) deferred func still twists my head. This whole reversed defer processing makes this really hard to understand.
I think the code is correct, but if there is a cleaner way to write this, the code would be more maintainable.

ubunatic (approved):
LGTM (the error handling could cleanup in another PR, if possible at all)

processor.go
frairon: I also don't like to modify the the return value in a deferred function, but haven't found a better solution yet. Maybe we'll find another way.

PR #354, opened 2021-10-03
Expose Processor State With StateReader Interface

### LINKED ISSUE #343: Propose: expose processor state
It will be nice to add a method to expose the processor.state field in order to have more fine grained information about processor state.
Examples could be:

`func (g *Processor) GetState() *Signal {
	return g.state
}`

`func (g *Processor) IsInState(state State) bool {
	return g.state.IsState(state)
}`

jomaresch:
Closes #343 
- add a `StateReader` to the processor to expose the current state.
- `PartitionTable.WaitRecovered()` now returns a read only channel.

### DIFF
--- a/partition_table.go
+++ b/partition_table.go
@@ -667,7 +667,7 @@ func (p *PartitionTable) CurrentState() PartitionStatus {
 }
 
 // WaitRecovered returns a channel that closes when the partition table enters state `PartitionRunning`
-func (p *PartitionTable) WaitRecovered() chan struct{} {
+func (p *PartitionTable) WaitRecovered() <-chan struct{} {
 	return p.state.WaitForState(State(PartitionRunning))
 }
 
--- a/signal.go
+++ b/signal.go
@@ -14,6 +14,12 @@ type waiter struct {
 	minState bool
 }
 
+// StateReader is a read only abstraction of a Signal to expose the current state.
+type StateReader interface {
+	State() State
+	WaitForState(state State) <-chan struct{}
+}
+
 // Signal allows synchronization on a state, waiting for that state and checking
 // the current state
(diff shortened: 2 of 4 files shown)


partition_table.go
```diff
@@ -667,7 +667,7 @@ func (p *PartitionTable) CurrentState() PartitionStatus {
 }
 
 // WaitRecovered returns a channel that closes when the partition table enters state `PartitionRunning`
-func (p *PartitionTable) WaitRecovered() chan struct{} {
+func (p *PartitionTable) WaitRecovered() <-chan struct{} {
 	return p.state.WaitForState(State(PartitionRunning))
 }
```
jomaresch, line 673: IMO this should be a read only channel.
But It's a breaking change. 
Don't know if we should keep it.

frairon: it's breaking but it fixes a bug so we definitely have to change that. If someone deliberately closed the channel from the outside it was wrong anyway :)

signal.go
```diff
@@ -14,6 +14,12 @@ type waiter struct {
 	minState bool
 }
 
+// StateReader is a read only abstraction of a Signal to expose the current state.
+type StateReader interface {
+	State() State
+	WaitForState(state State) <-chan struct{}
```
frairon, line 20: What about the other functions? I know they're mostly duplication, but it's convenient :). And if we have them anyway, why not expose them in the reader interface?
I mean `IsState` and `WaitForStateMin` and `ObserveStateChange`

ubunatic: Good idea + maybe we can change the places where we read/wait/watch state (if any) to use the reader and not access the state directly (?)

ubunatic (approved):
LGTM

Issue #160, opened 2018-12-06
Proposal: add tracing support


yuhui-lin:
Hey, I tried to add tracing for sarama and LevelDB using these two libraries:
https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/Shopify/sarama
https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/syndtr/goleveldb/leveldb

but I failed to do so because producer/consumer/leveldb.DB are not exposed directly. 
#77 adding header to goka context may be helpful but we still can't get traces of emitter/view/groupTable. 

I think it would be good to add opentracing support to the options of Emitter/processor/view:
https://github.com/opentracing/opentracing-go
```
func WithEmitterTracer(tracer opentracing.Tracer) EmitterOption
func WithProcessorTracer(tracer opentracing.Tracer) EmitterOption
func WithViewTracer(tracer opentracing.Tracer) ViewOption
```
any suggestion on how to add tracing to goka for now?

db7:
The storage is an interface, so you should be able to simply replace it with your extended version of the storage. Take a look in the [builders in `storage` subpackage](https://github.com/lovoo/goka/blob/21b04ba791d28c32ee85e8ef1682aea85d120b8c/storage/builders.go#L30). The same applies to the Sarama consumer and producer (`kafka` package).

If possible I'd suggest to introduce the tracing via builders instead of adding another dependency to the main package.

Sami has been working in a rewrite of the storage package. @SamiHiltunen, do you think that the rewrite would help introducing tracing?

yuhui-lin:
@db7 thanks for the information. I didn't realize they are all interfaces and I'll look into it. I guess it would still make sense if goka provides open-tracing support out of box.

db7:
Good points. I'd be more than happy reviewing PRs for this issue and #77. If you'd like to try, please let me know. I can also support your via slack.

frairon:
I'll close it for now, reopen or create a new issue if it comes up again.

adw1n:
I think there is definitely an interest in using tracing with goka. Right now to use https://godoc.org/gopkg.in/DataDog/dd-trace-go.v1/contrib/Shopify/sarama with a producer you need to implement a custom `ProducerBuiler` which involves copy pasting the whole `producer.go` file and changing just one line https://github.com/lovoo/goka/blob/3d39c5afcd0a0d967c49318208e2b313bc0dfc8c/producer.go#L32 to
```
[...]

PR #356, opened 2021-10-19
Allow wrapping Sarama producer/consumer for OpenTelemetry support

eko:
Hi @frairon,

I would like to implement [OpenTelemetry](https://opentelemetry.io/) support with Goka.

F25D98CE-6B67-4A9A-8C5F-0118EC2D75C7

I've seen that some discussions have already started on issue https://github.com/lovoo/goka/issues/160 some time ago.

In order to do this, I would like to use the [opentelemetry-go-contrib/Shopify/sarama](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/Shopify/sarama/otelsarama) instrumentation library because Goka is using it 🙂

## How it works

In order to implement it, I need to be able to wrap the Sarama producer / consumer so I've added some `ProcessorOptions` which works like this and does not modify the actual public signatures:

```go
opts := []goka.ProcessorOption{
	goka.WithSaramaProducerWrapper(func(producer sarama.AsyncProducer, config *sarama.Config) sarama.AsyncProducer {
		return otelsarama.WrapAsyncProducer(config, producer)
	}),

	goka.WithSaramaConsumerGroupHandlerWrapper(func(consumerGroupHandler sarama.ConsumerGroupHandler) sarama.ConsumerGroupHandler {
		return otelsarama.WrapConsumerGroupHandler(consumerGroupHandler)
	}),

	goka.WithSaramaConsumerWrapper(func(consumer sarama.Consumer) sarama.Consumer {
		return otelsarama.WrapConsumer(consumer)
	}),
}

p, err := goka.NewProcessor([]string{"localhost:9092"}, group, opts...)
```

## Side notes

As the Sarama OpenTelemetry implementation is using the `msg.Metadata` field to store its tracing span information and Goka was using it too, I've updated the Goka usage to don't use it because it was creating some issues.

Instead of using the `Metadata` field in Goka to store the `Promise` object, I've created a `sync.Map` and the promise is retrieved using the message pointer as key.

## Feedbacks

I would really appreciate your feedbacks about these changes and hope to see the OpenTelemetry available very soon in Goka ✌️

Thank you!

### DIFF
--- a/builders.go
+++ b/builders.go
@@ -8,25 +8,22 @@ import (
 )
 
 // ProducerBuilder create a Kafka producer.
-type ProducerBuilder func(brokers []string, clientID string, hasher func() hash.Hash32, producerWrapper ProducerWrapper) (Producer, error)
-
-// ProducerWrapper function allows to wrap a Sarama producer and return the wrapped Sarama producer.
-type ProducerWrapper func(producer sarama.AsyncProducer, config *sarama.Config) sarama.AsyncProducer
+type ProducerBuilder func(brokers []string, clientID string, hasher func() hash.Hash32) (Producer, error)
 
 // DefaultProducerBuilder creates a Kafka producer using the Sarama library.
-func DefaultProducerBuilder(brokers []string, clientID string, hasher func() hash.Hash32, producerWrapper ProducerWrapper) (Producer, error) {
+func DefaultProducerBuilder(brokers []string, clientID string, hasher func() hash.Hash32) (Producer, error) {
 	config := globalConfig
 	config.ClientID = clientID
 	config.Producer.Partitioner = sarama.NewCustomHashPartitioner(hasher)
-	return NewProducer(brokers, &config, producerWrapper)
+	return NewProducer(brokers, &config)
 }
 
 // ProducerBuilderWithConfig creates a Kafka consumer using the Sarama library.
 func ProducerBuilderWithConfig(config *sarama.Config) ProducerBuilder {
-	return func(brokers []string, clientID string, hasher func() hash.Hash32, producerWrapper ProducerWrapper) (Producer, error) {
+	return func(brokers []string, clientID string, hasher func() hash.Hash32) (Producer, error) {
 		config.ClientID = clientID
 		config.Producer.Partitioner = sarama.NewCustomHashPartitioner(hasher)
-		return NewProducer(brokers, config, producerWrapper)
+		return NewProducer(brokers, config)
 	}
 }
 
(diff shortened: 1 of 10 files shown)


frairon:
@eko , sorry for the long delay. The other error-handling PR really kept me busy for some time. 
Also very sorry for proposing yet another solution, but I have to :)
I thought about this whole wrapping solution and again, it feels like such a specific thing solved in such a generic way that it just doesn't look right. 
I mean if someone wants opentel support, they just want to add it easily, not apply the opentel wrappers using goka wrappers. Someone will forget to use 3 wrappers in processors and 1 for emitters and views and get errors and and and... you get the point, it is a bit complex.
Further, building a generic wrapper infrastructure assumes that there are more that work similar, but we don't have any indication that this is the case or will be. And I'm a bit afraid that it won't work without a lot of hacking to make it work with others. 

So what if we do not create any generic wrappers, but provide "tools" that can be used to tweak the components. I tried it in [this goka-tools-branch](https://github.com/lovoo/goka-tools/blob/opentel/opentel/options.go), which would allow you to create your components like this:

```go
// create options builder, optionally specifying different parent builders or sarama config,
// otherwise uses goka default.
otelbuilder := opentel.NewBuilder().WithConfig(...).WithConsumerBuilder(..)

goka.NewProcessor(...,
    otelbuilder.BuildProcOptions()...
)

goka.NewView(...,
    otelbuilder.BuildViewOptions()...
)

goka.NewEmitter(...,
    otelbuilder.BuildEmitterOptions()...
)
```

That way, goka would not be aware that it's being wrapped, so no major changes needed (except see notes). 
It's easy to apply, can be modified, and we hide the complex logic e.g. of wrapping the consumer-group-handler, which is actually the processor itself, behind a wrapper. 
I'd prefer that very much to having a random `if g.opts.consumerGroupHandlerWrapper != nil {...` in the middle of the rebalance-loop of the processor, which seems quite hacky and arbitrary.

## Notes 
If we do it that way, we will have to refactor the producer-builder to create the sarama-interface, and processors and emitters build their own wrapping `Producer` that handles the promises etc. 

Since the opentel-Producer-Wrapper seems to back up the metadata, we should be able to continue using this approach and avoid the sync-map. Otherwise it's a bug in the opentel-support which we should fix.

I know I said earlier to avoid touching the builder-interfaces. But seeing the producer-builder like this seems super inconsistent. So we'll have to live with the API change I guess :)

What are your thoughts on that?

eko:
Hi @frairon,

Thank you for working on this P.o.C.

I think you're right it could be better to provide a unified way to build the OpenTelemetry instrumentation for both producer and consumer / consumer group.

In case of a custom builder implementation users can still implement their own Builder implementation if they have some special cases like multiple wrapping so I think it's okay.

> Since the opentel-Producer-Wrapper seems to back up the metadata, we should be able to continue using this approach and avoid the sync-map. Otherwise it's a bug in the opentel-support which we should fix.

Concerning this metadata implementation, I won't say it's a bug but a wrong way to implement it ;-)

You (Goka) and the otelsarama instrumentation both use this field to maintain your state/promise but I think the issue is that you should not use it (if possible, and it is) because this field should be left to the end-user to use it. So I think both implementations have to be fixed.

What are the next step for you to have a first version available?

Thanks again for your time!

frairon:
Hi @korrawit  - wow that has been abandoned some time ago. To be honest I can't remember what the issues were exactly here, I think the main problem was how to get the instrumentation injected as painless as possible without messing with goka's API for too much. Let me check it out in more detail again and get back to you!

frairon:
@korrawit, I checked some things.
* sarama moved from shopify to IBM, for that the original otel-support was dropped.
* there is a new repo https://github.com/dnwe/otelsarama that does the same and it even improves on the metadata-usage, so the changes originally proposed in *this* PR are not necessary anymore.
* I'd still try to avoid adding another "wrapper"-layer proposed in this PR, instead we should provide support for otel as tools that use goka's options-interface. There was some work started in this [branch](https://github.com/lovoo/goka-tools/tree/opentel) , but for this we need to add a `SaramaProducerBuilder` option to to goka. That was already started back in the days in [this branch](https://github.com/lovoo/goka/compare/sarama-producer-builder?expand=1), but I think it doesn't work that way. Anyway, once we have this we can add the otel-tools either to `goka-tools` or into `goka` directly. 

So.. that's the plan I guess. What do you think?

Issue #350, opened 2021-09-28
View.Get fails on reconnecting view


frairon:
Goka views can be configured to tolerate Kafka connection issues and reconnect.

Currently, a view returns an error calling `.Get` when the view is in reconnecting state.

frairon:
fixed.

PR #361, opened 2021-11-07
Improved failure handling fully remove multierr.Errors system test for testing connection errors bugfix issue 350 View.Get fails on reconnecting view bugfix processor freeze on error (334)

frairon:
## Improve system tests
* replace flags-based trigger by env variable
* create proxy to test kafka-connection errors
* test processor disconnect
* test view reconnect

## improve error handling
* removing `multierr.Error` cosequently
* removing complexity when returning errors (e.g. removed some `errs.Collect` in `defer`ed functions)
* bugfix issue #350 View.Get fails on reconnecting view
* proper shutdown of processor on connection issues.

### DIFF
--- a/partition_table.go
+++ b/partition_table.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/Shopify/sarama"
+	"github.com/hashicorp/go-multierror"
 	"github.com/lovoo/goka/multierr"
 	"github.com/lovoo/goka/storage"
 )
@@ -17,6 +18,8 @@ const (
 
 	// internal offset we use to detect if the offset has never been stored locally
 	offsetNotStored int64 = -3
+
+	consumerDrainTimeout = time.Second
 )
 
 // Backoff is used for adding backoff capabilities to the restarting
@@ -191,7 +194,6 @@ func (p *PartitionTable) Close() error {
 func (p *PartitionTable) createStorage(ctx context.Context) (*storageProxy, error) {
 	var (
 		err   error
-		errs  = new(multierr.Errors)
 		st    storage.Storage
 		start = time.Now()
 		done  = make(chan struct{})
@@ -217,9 +219,7 @@ func (p *PartitionTable) createStorage(ctx context.Context) (*storageProxy, erro
 	}
 	err = st.Open()
 	if err != nil {
-		errs.Collect(st.Close())
-		errs.Collect(fmt.Errorf("error opening storage: %v", err))
-		return nil, errs.NilOrError()
+		return nil, multierror.Append(st.Close(), fmt.Errorf("error opening storage: %v", err)).ErrorOrNil()
 	}
 
 	// close the db if context was cancelled before the builder returned
@@ -228,10 +228,9 @@ func (p *PartitionTable) createStorage(ctx context.Context) (*storageProxy, erro
 		err = st.Close()
 		// only collect context error if Close() errored out
 		if err != nil {
-			errs.Collect(err)
-			errs.Collect(ctx.Err())
+			return nil, multierror.Append(err, ctx.Err()).ErrorOrNil()
 		}
-		return nil, errs.NilOrError()
+		return nil, nil
 	default:
 	}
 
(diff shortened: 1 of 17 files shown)


view.go
```diff
@@ -425,6 +430,12 @@ func (v *View) Evict(key string) error {
 
 // Recovered returns true when the view has caught up with events from kafka.
 func (v *View) Recovered() bool {
+	// no partitions --> never recover
+	// Otherwise we might mask errors of initializing the view
+	if len(v.partitions) == 0 {
```
frairon, line 435: possible bugfix: if views were empty due to errors, they might have appeared to be recovered.

view.go
```diff
@@ -331,6 +331,11 @@ func (v *View) Topic() string {
 // Get can be called by multiple goroutines concurrently.
 // Get can only be called after Recovered returns true.
 func (v *View) Get(key string) (interface{}, error) {
+
+	if v.state.IsState(State(ViewStateIdle)) || v.state.IsState(State(ViewStateInitializing)) {
```
frairon, line 335: block attempts to read if the view isn't ready yet.

partition_table.go
```diff
@@ -711,7 +673,7 @@ func (p *PartitionTable) IteratorWithRange(start []byte, limit []byte) (storage.
 
 func (p *PartitionTable) readyToRead() error {
 	pstate := p.CurrentState()
-	if pstate != PartitionRunning {
+	if pstate < PartitionConnecting {
```
frairon, line 676: This actually fixes bug #350

partition_table.go
```diff
@@ -424,43 +402,27 @@ func (p *PartitionTable) markRecovered(ctx context.Context) error {
...
-			return
-		}
-	}
-}
-
-func (p *PartitionTable) drainConsumer(cons sarama.PartitionConsumer, errs *multierr.Errors) {
+func (p *PartitionTable) drainConsumer(cons sarama.PartitionConsumer) error {
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
ubunatic, line 407: You changed the time constant. Maybe give it a name and list it as constant in the top of the file.

partition_table.go
```diff
@@ -480,30 +442,31 @@ func (p *PartitionTable) drainConsumer(cons sarama.PartitionConsumer, errs *mult
...
 
 	lastMessage := time.Now()
 
+	messages := cons.Messages()
+	errors := cons.Errors()
+
 	for {
 		select {
-		case msg, ok := <-cons.Messages():
+		case err, ok := <-errors:
```
ubunatic, line 460: I would use `more` instead of `ok` when checking if channels are open.
Then you can write
```
if !more {
    // stop processing
}
```

processor.go
```diff
@@ -352,33 +345,65 @@ func (g *Processor) rebalanceLoop(ctx context.Context) (rerr error) {
...
+func (g *Processor) handleSessionErrors(ctx, sessionCtx context.Context, sessionCtxCancel context.CancelFunc, consumerGroup sarama.ConsumerGroup) {
+	errs := consumerGroup.Errors()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-sessionCtx.Done():
+			return
+		case err, ok := <-errs:
```
ubunatic, line 381: `more` > `ok` :smile:

processor.go
```diff
@@ -352,33 +345,65 @@ func (g *Processor) rebalanceLoop(ctx context.Context) (rerr error) {
...
 }
 
+func (g *Processor) handleSessionErrors(ctx, sessionCtx context.Context, sessionCtxCancel context.CancelFunc, consumerGroup sarama.ConsumerGroup) {
+	errs := consumerGroup.Errors()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-sessionCtx.Done():
```
ubunatic, line 379: Either context.Done will break the loop. Is this correct or do you need to wait for both?

processor.go
frairon: in the next PR :)

processor.go
frairon: yeah, because `sessionCtx` recreated on each loop iteration and starts the error-reading go-routine again. But we need to stop this go-routine on rebalance (where there is no error), otherwise we end up with multiple of those routines. The parent context (ctx) however will not be cancelled yet, because there was not an error yet.

partition_table.go
frairon: yeah because I thought draining the channels into the void shouldn't take 10 seconds :). But good point, let's put it up as a constant.

frairon: done.

partition_table.go
```diff
@@ -424,43 +404,27 @@ func (p *PartitionTable) markRecovered(ctx context.Context) error {
...
 	errg.Go(func() error {
+		var errs *multierror.Error
+
 		for {
 			select {
-			case <-ctx.Done():
+			case <-timeoutCtx.Done():
 				p.log.Printf("draining errors channel timed out")
-				return nil
+				return errs
```
mmreza79, line 422: shouldn't return here `nil`?

frairon: i thought we should still return the errors we have collected so far, even if it times out.

mmreza79 (approved):

PR #363, opened 2021-11-17
Visitor/Tests imrovement

frairon:
This PR improves the processor's visiting-functionality:
* the call returns when all values have been visited
* the number of visited values are returned
* the visit stops upon rebalance or processor shutdown, notifying with a dedicated error

The PR also adds tests for visiting and improves other system tests in various places.

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -896,53 +896,50 @@ func (g *Processor) Stop() {
 	g.cancel()
 }
 
-// VisitAll visits all keys in parallel by passing the visit request
+// VisitAllWithStats visits all keys in parallel by passing the visit request
 // to all partitions.
-// An optional argument "meta" can be passed that will be forwarded to
-// the visit-function of each key of the table.
-// The function returns when a visit to all keys has been triggered or the context is cancelled.
-// Note that this does not imply that all processor callback have already processed the visit-request.
-//
-// TODO (fe): currently we're not triggering to abort a visit upon rebalance, because
-// a partition processor does not know when it's shutting down.
-// That implies that a rebalance will be blocked if there's a running visit request
-func (g *Processor) VisitAll(ctx context.Context, name string, meta interface{}) error {
+// The optional argument "meta" will be forwarded to the visit-function of each key of the table.
+// The function returns when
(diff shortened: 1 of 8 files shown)


processor.go
```diff
@@ -896,53 +897,50 @@ func (g *Processor) Stop() {
...
-		}
-		return nil
-	})
+	g.mTables.RUnlock()
 
-	return errg.Wait().ErrorOrNil()
+	// wait for the visitors
+	err := errg.Wait().ErrorOrNil()
+	return atomic.LoadInt64(&visited), err
+}
```
jomaresch, line 938: `atomic.LoadInt64(&visited)` will only return a number. 
This is a bit useless right? Because the caller want's to get the current value.
How about 
```
return func() {return atomic.LoadInt64(&visited)}, err
```

this way the caller can always "query" the current progess.

Or is there something i miss?

jomaresch: maybe this was also not the intention behind the whole `VisitAllWithStats` function

frairon: the atomic.Load is a bit confusing,because after `errg.Wait()` has returned, `visited` will not change anymore, so the atomic-call can be removed.
But the point is: VisitAllWithStats only returns after all visits are done, so no need to return a function.
We could, however, provide an option to pass a polling-function if we want, so the caller can poll for the progress while `Visit` runs (and blocks). Do you think that'd be necessary?

frairon: (On the other hand, if someone needs to know the progress, they can add some kind of metric in the visit-handler defined in the group graph)

jomaresch: ahh i missed the `err := errg.Wait().ErrorOrNil()` i thought `VisitAllWithStats` returns instantly after spawning all the go routines. Sorry.
But the we don't need `atomic.LoadInt64(&visited)` because nobody writes to `visited` anymore.

> (On the other hand, if someone needs to know the progress, they can add some kind of metric in the visit-handler defined in the group graph)

right

jomaresch (dismissed):

jomaresch (approved):

PR #366, opened 2021-12-07
hotfix: rebalance should not shut down processor

frairon:


processor.go
```diff
@@ -363,7 +363,7 @@ func (g *Processor) rebalanceLoop(ctx context.Context) (rerr error) {
 
 		select {
 		case <-time.After(5 * time.Second):
-		case <-sessionCtx.Done():
+		case <-ctx.Done():
```
frairon, line 366: Using the wrong context was shutting down the processor on eveyr rebalance, so there was essentially no rebalance.

juandes (approved):

PR #367, opened 2021-12-19
bugfix backoff: add max-wait, actually use parameter

frairon:
We use a simple backoff-mechanism for reconnecting views. That implementation had some isues that are fixed now:

* actually use the backoff-step-parameter
* add a max-wait parameter. Otherwise if the backoff would grow indefinitely and won't reconnect in reasonable time, when Kafka was offline for too long.

### DIFF
--- a/simple_backoff.go
+++ b/simple_backoff.go
@@ -4,15 +4,17 @@ import "time"
 
 // NewSimpleBackoff returns a simple backoff waiting the
 // specified duration longer each iteration until reset.
-func NewSimpleBackoff(step time.Duration) Backoff {
+func NewSimpleBackoff(step time.Duration, max time.Duration) Backoff {
 	return &simpleBackoff{
-		step: time.Second,
+		step: step,
+		max:  max,
 	}
 }
 
 type simpleBackoff struct {
 	current time.Duration
 	step    time.Duration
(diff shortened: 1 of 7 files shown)


simple_backoff.go
```diff
@@ -4,22 +4,27 @@ import "time"
...
 	b.current = time.Duration(0)
 }
 
 func (b *simpleBackoff) Duration() time.Duration {
-	b.current += b.step
-	return b.current
+	value := b.current
+	if b.current < b.max {
+		b.current += b.step
+	}
```
jomaresch, line 28: This is just a minor thing:
if `current = 100`, `step=20` and `max=110` and you run `Duration` `current` will be `120` which is larger than `max`. I think we should not exceed the max value.
```
if (b.current + b.step) <= b.max {
	b.current += b.step
}
```

jomaresch (dismissed):

simple_backoff.go
frairon: yeah true, but I thought "who cares if we'd wait a couple of seconds more. On the other hand you're right we should not exceed the limit, so let's change that.

frairon: fixed.

jomaresch (approved):

PR #373, opened 2022-03-06
github action for installing goka

frairon:
This PR improves testing of goka
* add action to install goka
* replace custom test functions with `testify`
* avoid stopping processors simultateously to avoid shutdown issues which would fail the test.


juandes (approved):

systemtest/processor_test.go
```diff
@@ -71,46 +70,48 @@ func TestHotStandby(t *testing.T) {
...
+	require.NoError(t, err)
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	errg, ctx := multierr.NewErrGroup(ctx)
 
 	errg.Go(func() error {
-		return proc1.Run(ctx)
+		// simulatenously stopping multiple processors sometimes fails the processors, so this one gets delayed
+		// see issue #376 for details
+		return proc1.Run(DelayedCtxCloser(ctx, 5*time.Second))
```
frairon, line 81: workaround to make the test more stable: avoid shutting down the processors simultaneously

systemtest/processor_test.go
```diff
@@ -141,44 +142,62 @@ func TestHotStandby(t *testing.T) {
...
+	cfg.Consumer.Offsets.Initial = sarama.OffsetOldest
+	goka.ReplaceGlobalConfig(cfg)
+
 	tm, err := goka.TopicManagerBuilderWithConfig(cfg, tmc)(brokers)
-	test.AssertNil(t, err)
+	require.NoError(t, err)
 
 	err = tm.EnsureStreamExists(inputStream, 1)
-	test.AssertNil(t, err)
+	require.NoError(t, err)
```
frairon, line 182: this test was flaky altogether, because it created the wrong topic, which ended up with two topics having two different number of partitions, as the examples-cluster was configured with "auto-create:true". This is fixed now and yields two new issues:
* fix all examples that depend on "auto-create:true" (#379)
* detect partition mismatches (#378)

jomaresch (approved):

Issue #99, opened 2018-03-05
Proposal: Implement windowing semantics


burdiyan:
It would be great if Goka could natively provide primitives for windowed processing. It’s a pretty complex topic, but seems really needed in many stream processing workloads. 

Kafka streams handles that by providing timestamp extractors to identify the actual time of the message, and modifying the record keys to be stored in the state store augmenting them with the window bucket.

Another approach that may help users to implement their own windowing is that if goka.Context would allow to get and set arbitrary values from the storage. I really don’t like it conceptually but it gives more freedom and is less complicated to implement.

I’m sure you were thinking about windowing in general. Do you have any clear idea on how it could be built into Goka?

db7:
This would be awesome! We don't have use cases for that at the moment, but I'm happy to discuss about it.

So `goka.Context` is an interface anyway. Whenever writing some value X, one could wrap that with some window bucket. The timestamp extractor could be either part of the codec or passed via option. One thing to consider is that automatically augmenting the state would make the table unreadable from views that are not aware of the augmentation.

Would you like to give a try? :-)

burdiyan:
I'd like to give it a try, but first would like to have it super clear in terms of the design.

>So goka.Context is an interface anyway. Whenever writing some value X, one could wrap that with some window bucket. 

The thing is that it seems like there's no way to swap the context implementation, it's somehow hardcoded in the processor here: https://sourcegraph.com/github.com/lovoo/goka/-/blob/processor.go#L643:9

What is the best way you see for someone to create custom context? Pass it as an option? Create something like context builder maybe?

>The timestamp extractor could be either part of the codec or passed via option.

I think both are valid solutions, but I think I like the option better.

>One thing to consider is that automatically augmenting the state would make the table unreadable from views that are not aware of the augmentation.

That's true. The thing is that in kafka streams nobody is supposed to use state store change log topics, they are only for restoring the state store. But Goka is explicitly allowing other pieces to use group table's changelog topic, which is also totally valid I think.

db7:
Nice!

I think it should be easy to wrap the callbacks in `goka.Input(topic, codec, callback)`. One idea would be to create a subpackage, say `timebucket`, then 
```go
g := tb.DefineGroup(
    tb.Input(topic, codec, callback), // callback will get wrapped
    tb.Persist(otherCodec), // otherCodec will get wrapped
)
````

And the wrapper would do something like this:
```
[...]

PR #392, opened 2022-08-23
Adds HSCAN key/value support to storage.redisIterator

chrisfjones:
The HSCAN result gives us a 1-dimensional array with [n] == key, [n+1] == value so we take advantage of that in the redisIterator.

FWIW We've been running this in production for the past 2 weeks with no issues. Before this change we were chugging through unexpected keys in `VisitAll`. Most of them were the *values* being passed as *keys*, a handful were the offset values (one per partition). After this change, we are only `Visit`ing the expected keys.

### DIFF
--- a/storage/redis/redis.go
+++ b/storage/redis/redis.go
@@ -133,6 +133,8 @@ func (s *redisStorage) Close() error {
 	return nil
 }
 
+// redisIterator is populated with the results of an HSCAN call for the table's key (https://redis.io/commands/scan/).
+// This result is a single-dimension array that contains [n] == key, [n+1] == value.
 type redisIterator struct {
 	current uint64
 	keys    []string
@@ -141,19 +143,26 @@ type redisIterator struct {
 }
 
 func (i *redisIterator) exhausted() bool {
-	return uint64(len(i.keys)) <= i.current
+	return uint64(len(i.keys)) <= i.current+1
 }
 
-func (i *redisIterator) Next() bool {
-	i.current++
-	if string(i.Key()) == offsetKey {
-		i.current++
+func (i *redisIterator) ignoreOffsetKey() bool {
+	if i.exhausted() {
+		return false
+	}
+	if string(i.keys[i.current]) == offsetKey {
+		i.current = i.current + 2
 	}
 	return !i.exhausted()
 }
 
+func (i *redisIterator) Next() bool {
+	i.current = i.current + 2
+	return i.ignoreOffsetKey()
+}
+
 func (i *redisIterator) Key() []byte {
-	if i.exhausted() {
+	if !i.ignoreOffsetKey() {
 		return nil
 	}
 	key := i.keys[i.current]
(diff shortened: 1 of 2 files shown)


chrisfjones:
@frairon 👋 
Let me know if there is more I should add to this PR. We've been running our fork in prod for about a month now and it's all working as expected 👍

frairon (approved):

frairon:
Hey @chrisfjones , 
many thanks for the contribution, sorry for the long delay - vacation took over :).
Anyway, to be honest we have not used the redis storage in a very long time - we can consider it unmaintained. Glad that you took a shot at improving it, really!! The changes don't look wrong, but I have no resources of double checking it with a running system at the moment, so let's just trust your prod environment.
Maybe we can add some system tests running as github actions in future.

One question out of curiosity: what's your use case for `VisitAll` here?
To iterate over a table, the recommended approach is `View.Iterator` as the `VisitAll` can really slow down a processor, so I'm just really curious.
Thanks again!

chrisfjones:
@frairon 👋 
No worries about the delay, hope you had a refreshing break!

My use case for `VisitAll` is that I've built a little "time-window aggregation" framework for our needs. It's intended to work a bit like https://docs.ksqldb.io/en/latest/concepts/time-and-windows-in-ksqldb-queries/ but a with a little more generic handling of "active sessions".

One of our use cases goes something like this (not exact, but it should illuminate the usage of goka):
We have a set of services that handle real-time collaborative text editing. For simplicity's sake, each character typed becomes an event. We want to "group" these events together by time and after, say, 15 seconds we want to emit a bundled event for further downstream processing. If the user types 1 character and then pauses for 15 seconds, we still want that bundled event to fire at the end.

To accomplish this in goka, we've done the following:
* For each "session" type we'd like to manage (typing bundle in this case), we create a goka group and an associated "SessionHandler". The input is the "all typing events" topic and the output is the "typing bundles" topic
* The SessionHandler has logic to decide if a session should start (always for this case) and if it should close (after 15 sec.)
* The state of a session is maintained in the group state.
* In the `process` function, we first see if a session is already open or if we need to create a new session.
* We then append whatever new state we want to for the session and see if we should close the session. 
* If the session is closed we delete the group state and `emit` the resulting bundle event to our output, otherwise we save our updated session as group state and move on.

This would work great on its own except for the case where the user types a few characters and then leaves the document. In this case, the "session" will remain open for that key until someone else decides to come in and type in the doc again. We chose to handle this case with `VisitAll` and it works like this:
* Every X seconds, we call `VisitAll` on all active keys for all partitions.
* We query their group state as a session and determine if that session can be closed (based on the current time)
* If the session is closed we delete the group state for that key and `emit` the resulting bundle event to our output.

Furthermore, we are using `RedisStorage` to keep track of our group state so that our process can handle process death and restarts without losing track of active sessions. So far we've adapted this pattern to various different session types running in production and although the `VisitAll` does consume a bit of processing, it seems to be working well for our needs.

Apologies for the tome 😄 and thanks for your time! Hopefully that makes sense. I'm 99% certain this solution is overkill and there is a much simpler way to accomplish this, so I'd be anxious to hear of any alternatives.

Cheers!

chrisfjones:
p.s. I've just now discovered #99 (a little late to the party 😅) and it looks like they're trying to accomplish a similar thing. I'll cross reference the above comment over there and would be happy to provide more detail on our solution.

frairon (approved):

frairon:
the two failing tests seem unrelated, let's merge this and fix the tests separately

PR #393, opened 2022-09-15
processor: add WaitForReadyContext method

tymeshifter:
Currently the only option to test for processor readines is `WaitForReady` method.
This method is blocking until processor is ready.

When processor is dealing with long initialization, this call becomes unresponsive for virtually indefinite time. This behaviour is deal breaker for monitoring calls that just want to check on the status of the processor and return in reasonable amount of time i.e. Kubernetes healthcheck.

This change adds context aware method `WaitForReadyContext` that has all the same checks with addition to context cancelation check. This allows it to return when context is canceled with the context reported error. If the error is not nil it indicates that processor is not ready.

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -734,22 +734,38 @@ func (g *Processor) Cleanup(session sarama.ConsumerGroupSession) error {
 	return err
 }
 
-// WaitForReady waits until the processor is ready to consume messages (or is actually consuming messages)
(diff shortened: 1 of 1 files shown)


frairon (approved):
looks good, thanks @tymeshifter for that! 
Any more comments from @jomaresch or @kavinsk1 maybe? :grin:

PR #397, opened 2022-10-13
Adds `Signal.WaitForStateMinWithCleanup` to avoid leaking waiters

chrisfjones:
If `Signal.WaitForState*` is called multiple times for a state like "stopping", the waiters will remain in the signal's queue forever and their channel will remain open. This has happened in my case while calling `PartitionProcessor.VisitValues` multiple times. This was causing `Signal.waiters` to grow indefinitely and cause a memory leak. 
This change adds a cleanup function to allow a caller to remove itself from the `Signal.waiters` list if it no longer cares about waiting for the given state.

fixes https://github.com/lovoo/goka/issues/398

### DIFF
--- a/signal.go
+++ b/signal.go
@@ -28,15 +28,18 @@ type StateReader interface {
 type Signal struct {
 	m                    sync.RWMutex
 	state                State
-	waiters              []*waiter
+	waiters              map[*waiter]struct{}
 	stateChangeObservers []*StateChangeObserver
 	allowedStates        map[State]bool
 }
+
+var emptyValue struct{}
 
 // NewSignal creates a new Signal based on the states
 func NewSignal(states ...State) *Signal {
 	s := &Signal{
 		allowedStates: make(map[State]bool),
+		waiters:       make(map[*waiter]struct{}),
 	}
 	for _, state := range states {
 		s.allowedStates[state] = true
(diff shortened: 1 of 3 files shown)


signal.go
```diff
@@ -107,6 +109,26 @@ func (s *Signal) WaitForStateMin(state State) <-chan struct{} {
 	return s.waitForWaiter(state, w)
 }
 
+// WaitForStateMinWithCleanup functions identically to WaitForStateMin, but returns a cleanup function in addition
+// so that the caller can cleanup resources if it no longer wants to wait
+func (s *Signal) WaitForStateMinWithCleanup(state State) (<-chan struct{}, func()) {
```
frairon, line 114: The name sounds a bit clumsy to be honest, almost like java :D, but I don't have a better idea and introducing breaking changes just for this seems overkill. The signal implementation never was the finest piece of software in the first place, so maybe let's just do it the way you proposed.

frairon (approved):
Thanks a lot for the fix!

signal.go
chrisfjones: Yeah, I struggled a bit with the naming 😆 
If there are more cases where this cleanup is needed down the road, the interface could be simplified and just always provide it.

chrisfjones: @frairon any idea what's up with the actions failures? It seems to be because it's running from my fork?

frairon: very sorry for the long delay. Let's just keep it that way and merge it. 
Also, don't worry about the actions, they're quite flaky right now and we'll try to take care of it.

frairon:
sorry again for the very long delay, I'll merge that now, many thanks for the fix!

PR #400, opened 2022-11-26
Fix clone functions of InputStats and OutputStats to prevent data race condition

leepro:
`&(*x)` doesn't clone the data (that pattern has used for the `clone()` function for `goka.InputStats` and `goka.OutputStats`) into a new instance because it is simplified to `x`. (This can be detected by [staticcheck SA4001](https://staticcheck.io/docs/checks#SA4001)) So, it has made data race condition when a cloned stats is used. We have observed this in Goka web monitor and other places where it is using `View.Stats`.

This can be reproduced with a goka web monitor enabled application by turning on the Go race detector `-race`.

Screen Shot 2022-11-26 at 12 51 42 AM

Screen Shot 2022-11-26 at 12 50 59 AM

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -93,11 +93,13 @@ func newOutputStats() *OutputStats {
 }
 
 func (is *InputStats) clone() *InputStats {
-	return &(*is)
+	var clone = *is
+	return &clone
 }
 
 func (os *OutputStats) clone() *OutputStats {
-	return &(*os)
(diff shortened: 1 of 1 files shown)


frairon:
good catch, many thanks for the fix!

frairon:
really sorry for the delay, things got out of sight over here.

PR #426, opened 2023-07-07
Fix race condition

zacharyfox:
Fixes a race condition in the partition table backoff / retry where multiple go routines try to access data due to being used in time.AfterFunc callbacks.

### DIFF
--- a/partition_table.go
+++ b/partition_table.go
@@ -3,6 +3,7 @@ package goka
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/Shopify/sarama"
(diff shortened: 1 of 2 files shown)


partition_table.go
```diff
@@ -137,27 +138,34 @@ func (p *PartitionTable) loadRestarting(ctx context.Context, stopAfterCatchup bo
 	var (
 		resetTimer *time.Timer
 		retries    int
+		retryMux   sync.Mutex
```
frairon, line 141: instead of the mutex, we could use `atomic.Int32`, do you think that would improve readability?

frairon (commented):
Many thanks for the contribution and the PR! I think there's nothing to complain here, just one question :)

enzshew (approved):

frairon (approved):

Issue #433, opened 2023-09-19
"panic:  sync: negative WaitGroup counter" in PartitionProcessor.VisitValues


akshatraika-moment:
Hi, seeing a panic loop in https://github.com/lovoo/goka/blob/master/partition_processor.go#L685  and I am using version 1.1.7

Here is the stack trace:
```
/go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:72 +0xa5
created by golang.org/x/sync/errgroup.(*Group).Go
/go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:75 +0x64
golang.org/x/sync/errgroup.(*Group).Go.func1()
/go/pkg/mod/github.com/lovoo/goka@v1.1.7/multierr/errgroup.go:48 +0x52
github.com/lovoo/goka/multierr.(*ErrGroup).Go.func1()
/go/pkg/mod/github.com/lovoo/goka@v1.1.7/processor.go:934 +0x39
github.com/lovoo/goka.(*Processor).VisitAllWithStats.func1()
/go/pkg/mod/github.com/lovoo/goka@v1.1.7/partition_processor.go:716 +0x69a
github.com/lovoo/goka.(*PartitionProcessor).VisitValues(0xc000a20500, {0x1607528, 0xc0178d4c40}, {0x1270c8a, 0x8}, {0x1124020?, 0xc00062d560}, 0xc0188d0550)
/go/pkg/mod/github.com/lovoo/goka@v1.1.7/partition_processor.go:701
github.com/lovoo/goka.(*PartitionProcessor).VisitValues.func1(...)
/usr/local/go/src/sync/waitgroup.go:108
sync.(*WaitGroup).Done(...)
/usr/local/go/src/sync/waitgroup.go:83 +0xda
sync.(*WaitGroup).Add(0xc018024e48?, 0xc018024d7c?)
goroutine 1087 [running]:
panic: sync: negative WaitGroup counter
```

There is possibly a race condition in the VisitValues function. Can someone please take a look?

Happy to provide more information if needed but this seems to be independent of my implementation.

mikemoment:
Can someone please take a look at this issue? It caused a failure in our production env. Thanks!

asoliman20:
+1 

I'm also having this issue!

frairon:
Could you provide some information how you use the visit functionality? Like how is it initialized, what's done inside the visit function and so on?

[...]

PR #434, opened 2023-09-22
issue #433: bugfix on message drain for panic in visits

frairon:
Should fix issue #433

### DIFF
--- a/systemtest/processor_visit_test.go
+++ b/systemtest/processor_visit_test.go
@@ -13,6 +13,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// size of the channel used for the visitor, defined in goka/config.go
+var (
+	visitChannelSize = 100
+	numPartitions    = 10
+)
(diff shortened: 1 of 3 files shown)


akshatraika-moment (approved):

systemtest/processor_visit_test.go
```diff
@@ -128,6 +134,87 @@ func TestProcessorVisit(t *testing.T) {
...
+		})
+
+		visitCtx, visitCancel := context.WithCancel(context.Background())
+		defer visitCancel()
+
+		var (
+			visitErr  error
+			visitDone = make(chan struct{})
+		)
+		// pass wrong type to visitor -> which will be passed to the visit --> will panic
```
norbertklawikowski, line 201: Comment is not correct because we pass the correct type -> should be removed.

frairon: alright, fixed it.

PR #442, opened 2023-11-20
Correct README's example usage

danielgospodinow:
For the usage example in the README it's mentioned that the emitter runs once and stops while it actually runs forever and keeps emitting a message every second.

We can just correct that single statement.


frairon (approved):
thanks a lot

PR #449, opened 2024-04-06
Improve processor stopped state management

jalaziz:
* Introduce a new `Stopped` state for processors. Since a processor cannot be restarted after it has been run, this state better indicates that the processor has reached its final state.
* Expose a new `Done()` method on the processor to allow waiting on a processor to complete. This avoids the need to apply additional "done" channels on top of the processor when processors in go routines.

### DIFF
--- a/processor_test.go
+++ b/processor_test.go
@@ -351,7 +351,7 @@ func TestProcessor_Run(t *testing.T) {
 		consBuilder, _ := createTestConsumerBuilder(t)
 
 		graph := DefineGroup("test",
-			// not really used, we're failing anyway
+			// not really used, we're stopping before the processor before producing
 			Input("input", new(codec.Int64), accumulate),
 		)
 
--- a/processor_test.go
+++ b/processor_test.go
@@ -315,6 +315,7 @@ func TestProcessor_Run(t *testing.T) {
 		// and waiting for them to be delivered
 		<-done
 		require.True(t, strings.Contains(procErr.Error(), "setup-error"))
+		require.True(t, strings.Contains(newProc.Error().Error(), "setup-error"))
 	})
 }
 
@@ -410,12 +411,9 @@ func TestProcessor_Stop(t *testing.T) {
 			bm.createProcessorOptions(consBuilder, groupBuilder)...,
 		)
 		require.NoError(t, err)
-		var (
-			procErr error
-		)
 
 		go func() {
-			procErr = newProc.Run(ctx)
+			newProc.Run(ctx)
 		}()
 
 		newProc.WaitForReady()
@@ -425,7 +423,7 @@ func TestProcessor_Stop(t *testing.T) {
 
 		select {
 		case <-newProc.Done():
-			require.NoError(t, procErr)
+			require.NoError(t, newProc.Error())
 		case <-time.After(10 * time.Second):
 			t.Errorf("processor did not shut down as expected")
 		}
--- a/processor_test.go
+++ b/processor_test.go
@@ -10,9 +10,10 @@ import (
 
 	"github.com/IBM/sarama"
 	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+
 	"github.com/lovoo/goka/codec"
 	"github.com/lovoo/goka/storage"
-	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


processor_test.go
```diff
@@ -324,3 +325,111 @@ func TestProcessor_StateReader(t *testing.T) {
 
 	require.Equal(t, ProcStateRunning, p.StateReader().State())
 }
+
+func TestProcessor_Stop(t *testing.T) {
+	t.Run("expected-state", func(t *testing.T) {
```
frairon, line 330: If we do define subtests, then we could reduce to one top-level testing and use both `t.Run`s in side it. 

Like
```go

func TestProcessor_Stop(t*testing.T){
  t.Run("expected-state", func(t*testing.T){
   // ...
  }

  t.Run("done-closes", func(t *testing.T) {
   // ....
  })
}
```

Does that make sense?

frairon (commented):
Thanks  for your contribution  @jalaziz, sorry for the delay!

frairon:
:smile: the deadlock is removed, but the test reports a data-race.

processor_test.go
jalaziz: Yup! I went with different tests because we're technically testing different methods, but will combine them into one top level test 🙏🏼

jalaziz:
@frairon So actually, there _is_ a data race.

The data race happens because the channel will be closed _before_ the error is returned from `Run`.

I could fix the test by introducing an extra done channel for the text. However, another way to fix this is to have the `done` channel return the error that `Run` returns. This is kind of nice as you can handle the error asynchronously.

What do you think?

processor_test.go
frairon: yeah splitting would be fine too, just having two top levels with each one sub-test looks strange. So either combine or remove the subtest.

jalaziz: Ah fair, I've combined them into one function for now.

frairon:
True that, there is a data race, but the data race is in the test only. If we add a synchronizing channel or a mutex it's fixed. But of course, that's not optimal. 

On the other hand, returning the error through the done-channel also doesn't work, because only the first consumer might get the error. As soon as multiple clients wait for the processor it becomes inconsistent. We could create a new errors-channel each time the method is called and copy the error, which is even uglier. 

So what if we instead provide an `Error()` method on the processor that returns the last error? Clients could call when either `Run()` returned or the channel from `Done()` is closed, if they call it earlier - they'll get a race-error in their tests. Or we protect it with a mutex to be sure.

What do you think about that?

jalaziz:
@frairon I have implement what you suggested. Let me know what you think.

I'm a bit torn on whether all the tests should be updated to use `Done()` and `Error()`. Happy to update the tests if you think they should.

frairon:
@jalaziz thanks for this, really! Since we have the `Done()`-method now, I think it does make sense to use it in the tests. I mean why do the dance with the `defer close(done)` everywhere if you don't have to. But I'd understand if you don't want to change all the tests. But if you do - be my guest :).

jalaziz:
@frairon Updated the tests, but I added two tests to make sure the tests still covered the returned error.

frairon (dismissed):

frairon:
Looks good, thanks! @norbertklawikowski what do you think?

processor_test.go
```diff
@@ -315,6 +339,46 @@ func TestProcessor_Run(t *testing.T) {
...
+
+		bm.tmgr.EXPECT().Close().Times(1)
+		bm.tmgr.EXPECT().Partitions(gomock.Any()).Return([]int32{0}, nil).Times(1)
+		bm.producer.EXPECT().Close().Times(1)
+
+		groupBuilder, _ := createTestConsumerGroupBuilder(t)
+		consBuilder, _ := createTestConsumerBuilder(t)
+
+		graph := DefineGroup("test",
+			// not really used, we're failing anyway
```
norbertklawikowski, line 354: The comment is not valid for this test, since we expect no failure.

norbertklawikowski:
LGTM, just a tiny comment :)

processor_test.go
jalaziz: Reworded to be more accurate.

norbertklawikowski (approved):
Nice, thanks a lot!

PR #451, opened 2024-04-25
fix a deadlock in the consumer mock

moonpolysoft:
Found a deadlock in the `MockAutoPartitionConsumer`. The `Close()` function does all of its work from within a `sync.Once`. However, the `AsyncClose()` function utilizes the same `sync.Once` to do its work. `Close()` calls `AsyncClose()` from within the once invocation and results in a deadlock as the mutex is being held by the outer `Close()`.

This change inlines the operations `AsyncClose()` would've performed into the main body of `Close()`, removing the possibility for deadlock.


frairon (approved):

frairon:
Good catch, thanks a lot @moonpolysoft !

PR #453, opened 2024-05-06
Add message timestamp to view update context

smirzaei:
This PR adds the record timestamp to view's [UpdateContext](https://github.com/lovoo/goka/blob/85b1720e021673fd117d7a4aa507db488a148cf2/options.go#L16-L31).

Having access to record's timestamp helps us greatly with measuring the lag when the compacted topic is replicated in different regions.

I ran this version locally and it's working well. Please let me know if I missed something. Thank you!


frairon (approved):
@smirzaei thanks a lot, looks fine by me! @norbertklawikowski what do you think?

PR #455, opened 2024-07-15
Add options for librdkafka partitioning compatibility

owenniles:
This PR allows developers to configure Goka's partitioning behavior to match librdkafka. Open to feedback on the approach!

**Why?**
To allow the group table to be copartitioned with input topics whose producers use librdkafka's partitioning behavior. There is a subtle difference in the way librdkafka and Sarama convert hashes of keys to partition numbers (embarrassingly, [here is my source](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#SaramaCompatHasher) for this claim). If a Goka processor processing an input topic with librdkafka-based producers uses sarama's partitioning behavior, the group table will not be copartitioned with the input topics.

**Changes**
- Added a `goka.ProducerBuilder` constructor that allows developers more granular control over the producer's `sarama.HashPartitioner`, which is responsible for assigning keys to partitions.
  - Technically, you can already customize the `sarama.HashPartitioner` by writing a custom `goka.ProducerBuilder`, but I believe it would be nicer if Goka provided a first-party way to do this.
- Added a view option `goka.WithViewHashUnsigned` that configures the view to calculate the number of the partition in which a key can be found in the same way librdkafka would.

### DIFF
--- a/view.go
+++ b/view.go
@@ -295,6 +295,11 @@ func (v *View) close() error {
 }
 
 func (v *View) hash(key string) (int32, error) {
+	numPartitions := len(v.partitions)
+	if numPartitions < 1 {
+		return 0, errors.New("no partitions found")
+	}
+
 	// create a new hasher every time. Alternative would be to store the hash in
 	// view and every time reset the hasher (ie, hasher.Reset()). But that would
 	// also require us to protect the access of the hasher with a mutex.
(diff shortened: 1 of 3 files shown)


view.go
owenniles, line 1: The changes in this file were inspired by the implementation of [this function](https://github.com/IBM/sarama/blob/d2246cc9b9df113f3ffed7c442fd81e8b9cf736e/partitioner.go#L209).

owenniles:
It occurred to me that setting the partitioning behavior for each lookup table may be desirable as well https://github.com/lovoo/goka/issues/456.

PR #469, opened 2025-02-16
fix: Kafka client resource leak

jalaziz:
Sarama Kafka clients leak several resources (goroutines, metrics, connections) if not closed properly. A client is created and managed within the topic manager and therefore a topic manager will leak resources if not properly closed.

Topic Managers are created in two places:

1. When calling `prepareTopics` while instantiating a new processor.
2. Within `processor.Run`.

The first of these is correctly closed, but the latter was not. Closing the second instance should ensure underlying Sarama resources are appropriate cleaned up when a processor stops.

### DIFF
--- a/processor.go
+++ b/processor.go
@@ -288,6 +288,11 @@ func (g *Processor) Run(ctx context.Context) (rerr error) {
 	if err != nil {
 		return fmt.Errorf("Error creating topic manager for brokers [%s]: %v", strings.Join(g.brokers, ","), err)
 	}
+	defer func() {
+		if err := g.tmgr.Close(); err != nil {
+			errs = multierror.Append(errs, fmt.Errorf("error closing topic manager: %w", err))
+		}
+	}()
 
 	// create kafka producer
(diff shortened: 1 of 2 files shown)


frairon (approved):
awesome, thanks!

mmreza79 (approved):
nice :)

jalaziz:
Happy to help! If there's anything else I can do to help get this over the line and released, please let me know 🙏🏼. It was actively affecting some of our workloads, so we'd love to upgrade to a patched version.

jalaziz:
Just checking if there's a planned release with this fix in the near future? Don't mean to rush a release, just trying to figure out if we should wait for a release or publish or own patched version.

norbertklawikowski:
@jalaziz Let me check back with @frairon to see if we have everything for the next release, will notify you if we have released a new version.

norbertklawikowski:
@jalaziz FYI fix is contained in this release: https://github.com/lovoo/goka/releases/tag/v1.1.13 Thanks again for your help!

PR #473, opened 2025-02-25
improve processor stability

frairon:
## What this PR tries to improve
Stability of processors in the face of restarting/unstable kafka processors.

## Background
We're facing the issue that our kafka-cluster restarts or rebalances from time to time, which makes all processors restart. Since the processors will rebalance, this PR uses reconnecting views to be used for the join/lookup tables.

### DIFF
--- a/partition_processor.go
+++ b/partition_processor.go
@@ -257,7 +257,7 @@ func (pp *PartitionProcessor) Start(setupCtx, ctx context.Context) error {
 		join := join
 		pp.runnerGroup.Go(func() error {
 			defer pp.state.SetState(PPStateStopping)
-			return join.CatchupForever(runnerCtx, false)
+			return join.CatchupForever(runnerCtx, true)
 		})
 	}
 
@@ -274,7 +274,7 @@ func (pp *PartitionProcessor) Start(setupCtx, ctx context.Context) error {
 			// (b) run the processor table in catchup mode so it keeps updating it's state.
 		case runModePassive:
 			if pp.table != nil {
-				err = pp.table.CatchupForever(runnerCtx, false)
+				err = pp.table.CatchupForever(runnerCtx, true)
 			}
 		default:
 			err = fmt.Errorf("processor has invalid run mode")
@@ -298,15 +298,15 @@ func (pp *PartitionProcessor) Stop() error {
 	pp.state.SetState(PPStateStopping)
 	defer pp.state.SetState(PPStateStopped)
 
-	close(pp.input)
-	close(pp.visitInput)
-
 	if pp.cancelRunnerGroup != nil {
 		pp.cancelRunnerGroup()
 	}
 
 	// wait for the runner to be done
 	runningErrs := multierror.Append(pp.runnerGroup.Wait().ErrorOrNil())
+
+	close(pp.input)
+	close(pp.visitInput)
 
 	// close all the tables
 	stopErrg, _ := multierr.NewErrGroup(context.Background())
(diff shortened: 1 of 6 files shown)


partition_processor.go
```diff
@@ -257,7 +257,7 @@ func (pp *PartitionProcessor) Start(setupCtx, ctx context.Context) error {
 		join := join
 		pp.runnerGroup.Go(func() error {
 			defer pp.state.SetState(PPStateStopping)
-			return join.CatchupForever(runnerCtx, false)
+			return join.CatchupForever(runnerCtx, true)
```
frairon, line 260: this makes the join-table trying to reconnect while the processor is running (together with the other table some lines below)

partition_processor.go
```diff
@@ -298,16 +298,16 @@ func (pp *PartitionProcessor) Stop() error {
...
-	close(pp.visitInput)
-
 	if pp.cancelRunnerGroup != nil {
 		pp.cancelRunnerGroup()
 	}
 
 	// wait for the runner to be done
 	runningErrs := multierror.Append(pp.runnerGroup.Wait().ErrorOrNil())
 
+	close(pp.input)
```
frairon, line 308: channels are now closed after the runner-group is done --> visitors are attaching to the runner-group for this.

partition_processor.go
```diff
@@ -637,15 +637,6 @@ func (pp *PartitionProcessor) VisitValues(ctx context.Context, name string, meta
 
 	var wg sync.WaitGroup
 
-	// drains the channel and drops out when closed.
```
frairon, line 640: there was actually no point to distinguish between draining until close or draining until it's empty, because *this* function is writing to the channel.
In case two visitors are started at the same time and one of them panics or is stopped, it'll drain the other's messages too - but that is an issue that existed before, so we'll ignore it here :)

processor.go
```diff
@@ -421,11 +421,11 @@ func (g *Processor) handleSessionErrors(ctx, sessionCtx context.Context, session
 			)
 
 			if errors.As(err, &errProc) {
-				g.log.Debugf("error processing message (non-transient), shutting down processor: %v", err)
+				g.log.Printf("error processing message (non-transient), shutting down processor: %v", err)
```
frairon, line 424: let's have those important errors not as debug.

topic_manager.go
```diff
@@ -89,7 +89,7 @@ func checkBroker(broker Broker, config *sarama.Config) error {
 	}
 
 	err := broker.Open(config)
-	if err != nil {
+	if err != nil && !errors.Is(err, sarama.ErrAlreadyConnected) {
```
frairon, line 92: accordin to docs, `Open` might return this if it's already connected and it's not an error.

systemtest/proc_disconnect_test.go
```diff
@@ -18,6 +18,7 @@ func TestProcessorShutdown_KafkaDisconnect(t *testing.T) {
 	brokers := initSystemTest(t)
 	var (
 		topic = goka.Stream(fmt.Sprintf("goka_systemtest_proc_shutdown_disconnect-%d", time.Now().Unix()))
+		join  = goka.Stream(fmt.Sprintf("goka_systemtest_proc_shutdown_disconnect-%d-join", time.Now().Unix()))
```
frairon, line 21: adding some join tables to the tests so we can test the reconnecting joins change from above.

mmreza79 (approved):

norbertklawikowski (approved):

PR #477, opened 2025-04-25
fix displaying topic name

mmreza79:
When topic manger set `NoCreate` option it throws the following error message if the topic is not exist in kafka even though the topic name is exist. 
```
error creating goka processor: topic does not exist but the manager is configured with NoCreate, 
so it will not attempt to create it
```
May be for some reason it could not parse the topic name. Quoting the topic name displays the missing topic name.

### DIFF
--- a/topic_manager.go
+++ b/topic_manager.go
@@ -172,17 +172,15 @@ func (m *topicManager) ensureExists(topic string, npar, rfactor int, config map[
 
 	partitions, err := m.Partitions(topic)
 
-	if err != nil {
-		if err != errTopicNotFound {
-			return fmt.Errorf("error checking topic: %v", err)
-		}
+	if err != nil && err != errTopicNotFound {
+		return fmt.Errorf("error checking topic: %v", err)
 	}
 	// no topic yet, let's create it
 	if len(partitions) == 0 {
 
 		// (or not)
 		if m.topicManagerConfig.NoCreate {
-			return fmt.Errorf("topic `%s` does not exist but the manager is configured with NoCreate, so it will not attempt to create it", topic)
+			return topicNoCreateError(topic)
 		}
 
 		return m.createTopic(topic,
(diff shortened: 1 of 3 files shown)


topic_manager.go
```diff
@@ -182,7 +182,7 @@ func (m *topicManager) ensureExists(topic string, npar, rfactor int, config map[
 
 		// (or not)
 		if m.topicManagerConfig.NoCreate {
-			return fmt.Errorf("topic %s does not exist but the manager is configured with NoCreate, so it will not attempt to create it", topic)
+			return fmt.Errorf("topic `%s` does not exist but the manager is configured with NoCreate, so it will not attempt to create it", topic)
```
paulhenke, line 185:
```go
return topicNotFoundError(topic)
```

with
```go
type topicNotFoundError string

func (e topicNotFoundError) Error() string {
	return fmt.Sprintf("topic '%s' does not exist but the manager is configured with NoCreate, so it will not attempt to create it", string(e))
}
```

topic_manager_test.go
```diff
@@ -269,7 +269,7 @@ func TestTM_EnsureStreamExists(t *testing.T) {
 		)
 
 		err := tm.EnsureStreamExists(topic, npar)
-		require.ErrorContains(t, err, "will not attempt to create it")
+		require.Equal(t, err.Error(), "topic `some-topic` does not exist but the manager is configured with NoCreate, so it will not attempt to create it")
```
paulhenke, line 272:
```suggestion
require.True(t, errors.Is(err, topicNotFoundError("some-topic")))
```

topic_manager.go
paulhenke: I don't think it's better, but i also don't think it's worse.
You see the error from the type name. The message itself is IMO secondary.

https://github.com/lovoo/goka/pull/477/commits/160d15d3c3875110dffb3a9f90346c38e30ff430
This is not idiomatic Go. I think it's very ugly.
And usually we don't compare strings, because they're brittle. Instead we compare types, because changing the name of the type changes it everywhere.

mmreza79: alright :)

frairon (approved):

PR #485, opened 2026-03-10
fix the issue with offsetOutOfRange

dethlex:
**Summary**
Handle ErrOffsetOutOfRange in goka's partition table recovery by falling back to sarama.OffsetOldest instead of crashing the processor
                                                                                                                                                                                                                                                                                       
**Problem**
```                                                                                                                                                
  Production error:                                                                                                                                                                                                                                                        
  error consuming %TOPIC%: kafka server: The requested offset is outside   
  the range of offsets maintained by the server for the given topic/partition         
```
Goka's findOffsetToLoad() adjusts offsets below the broker's oldest, but a race condition exists: the oldest offset can shift between the metadata fetch and the actual ConsumePartition call (e.g. due to log compaction or retention cleanup), causing ErrOffsetOutOfRange.
                                                                                                                                                                                                                                                                                       
**Fix**                                                                                                                                                                                                                                                       When ConsumePartition returns ErrOffsetOutOfRange, retry with sarama.OffsetOldest (lets the broker resolve the actual earliest offset) instead of failing the partition setup.

### DIFF
--- a/partition_table.go
+++ b/partition_table.go
@@ -2,6 +2,7 @@ package goka
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -330,7 +331,7 @@ func (p *PartitionTable) load(ctx context.Context, stopAfterCatchup bool) (rerr 
 	defer p.log.Debugf("... Loading done")
 
 	partConsumer, err = p.consumer.ConsumePartition(p.topic, p.partition, loadOffset)
-	if err == sarama.ErrOffsetOutOfRange {
+	if errors.Is(err, sarama.ErrOffsetOutOfRange) {
 		p.log.Printf("Offset %d out of range for topic %s, partition %d. Falling back to oldest offset.", loadOffset, p.topic, p.partition)
 
 		partConsumer, err = p.consumer.ConsumePartition(p.topic, p.partition, sarama.OffsetOldest)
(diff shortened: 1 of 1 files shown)


partition_table.go
```diff
@@ -329,8 +331,13 @@ func (p *PartitionTable) load(ctx context.Context, stopAfterCatchup bool) (rerr
 	defer p.log.Debugf("... Loading done")
 
 	partConsumer, err = p.consumer.ConsumePartition(p.topic, p.partition, loadOffset)
-	if err != nil {
-		return fmt.Errorf("Error creating partition consumer for topic %s, partition %d, offset %d: %v", p.topic, p.partition, storedOffset, err)
+	if errors.Is(err, sarama.ErrOffsetOutOfRange) {
```
norbertklawikowski, line 334: The handling of other errors than `sarama.ErrOffsetOutOfRange` is missing here.

partition_table.go
```diff
@@ -347,6 +354,22 @@ func (p *PartitionTable) load(ctx context.Context, stopAfterCatchup bool) (rerr
 
 	// load messages and stop when you're at HWM
 	loadErr := p.loadMessages(ctx, partConsumer, hwm, stopAfterCatchup)
+	if loadErr != nil && errors.Is(loadErr, sarama.ErrOffsetOutOfRange) {
```
norbertklawikowski, line 357: Is it possible that a `ErrOffsetOutOfRange` can be returned by the `Errors` channel of the consumer inside `loadMessages`? It seems to be only possibly returned via the `chooseStartingOffset` which is called via `ConsumePartition`, thus while constructing a `PartitionConsumer`.

Here it would mean that while consuming from a `PartitionConsumer` this error can happen, and we would retry only one time which doesn't seem to be sufficient in this case. But I can not see that this is possible at all.

norbertklawikowski (commented):
Thanks for creating the PR!

PR #491, opened 2026-08-03
TSI-9010: update go version, docker compose v2 and run gofix

mmreza79:
GitHub Actions runners now ship with Docker Compose [V2](https://github.com/digorgonzola/github-runner-docker) as a plugin. Therefore, we need to use docker compose V2 to run docker containers for the examples.

We also run [gofix](https://go.dev/blog/introducing-gofix) in order to update existing code.


frairon (dismissed):

paulhenke:
run `go fix ./...` 
it will replace `interface{}` with `any`  and many more useful changes

https://go.dev/blog/gofix

examples/Makefile
```diff
@@ -17,8 +17,8 @@ stop: compose-down
 
 compose-up:
 	@echo start docker compose
-	@docker-compose up -d
+	@docker compose up -d
```
norbertklawikowski: Would also be nice to support podman here like in `wait-kafka.sh`.

frairon (approved):

